PR #170, opened 2020-03-30
Update Account Type w/ ARR Logic

rbradleyhaas:


transform/sql/account_type.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
+
+UPDATE orgm.account
+SET type = 'Customer (Attrited)'
+WHERE COALESCE(arr_current__c,0) = 0
+    AND type = 'Customer';
```
rbradleyhaas: @michaelschiffmm I don't want to do `type != 'Customer (Attrited)'` because then all prospects would be set to it too.

PR #250, opened 2020-05-21
Add more names to the staff list

jasonblais:
Wondering if we need to sync against Bamboo at some point to ensure all past and future staff are included in this list?


rbradleyhaas:
@jasonblais I have added these to the script that I run and populates this table. I am going to close this.

PR #335, opened 2020-07-13
Rudderstack Dev & RC Testing Models and Macros

enelson720:


enelson720:
These are the code changes that allow us to build dev and rc data models in analytics database. It involved the use of jinja to build new macros that enable us to union all tables across multiple schemas, as well as each data model for the various functional groupings of telemetry data.

PR #535, opened 2020-11-20
Add previous staff to GitHub contributor exclusion list

jasonblais:
@rbradleyhaas I wonder if we can include this in our offboarding process, since they are affecting community stats?


rbradleyhaas (approved):

rbradleyhaas:
@jasonblais the reason it does this is the current script overwrites based on the current staff members. I have tried to not run updates on this as much as possible.

PR #826, opened 2021-09-23
Revert sleep

nickmisasi:

### DIFF
--- a/billing/run_nightly_billing.py
+++ b/billing/run_nightly_billing.py
@@ -30,7 +30,7 @@ def main():
         blapi_token = os.getenv("BLAPI_TOKEN")
         blapi_url = os.getenv("BLAPI_URL")
         db_url = os.getenv("BLAPI_DATABASE_URL")
-    time.sleep(10000)
+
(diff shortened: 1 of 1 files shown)


billing/run_nightly_billing.py
```diff
@@ -30,7 +30,7 @@ def main():
         blapi_token = os.getenv("BLAPI_TOKEN")
         blapi_url = os.getenv("BLAPI_URL")
         db_url = os.getenv("BLAPI_DATABASE_URL")
-    time.sleep(10000)
+
```
neallred, line 33: Can we also remove the `import time` statement at the top?

neallred (approved):
lgtm! For my own learning, what was waiting supposed to achieve?

nickmisasi:
> lgtm! For my own learning, what was waiting supposed to achieve?

This code gets run inside its own pod, but was claiming success after a few seconds (where it didn't appear to do anything). We needed the sleep in order to keep the pod alive so we could poke around inside it to debug an issue

marianunez (approved):

PR #852, opened 2021-10-14
Warehousing

enelson720:


transform/snowflake-dbt/models/events/hourly/rudder_webapp_events.sql
```diff
@@ -2,7 +2,7 @@
     "materialized": "incremental",
     "schema": "events",
     "tags":"preunion",
-    "snowflake_warehouse": "transform_l",
```
catalintomai, line 5: @sri-byte , what is the difference between `transform_l` and `transform_m` ? Thanks.

sri-byte: Good question, Its the type of warehouse being used to process data, every warehouse has certain amount of credits and we pay snowflake based on credits used. We would be using more credits to use `transform_l `large vs. `transform_m `meduim. I believe this table got dropped accidentally when the user was trying to optimize the warehouse from large to medium.

https://docs.snowflake.com/en/user-guide/warehouses-overview.html

sri-byte: https://mattermost.looker.com/dashboards/47?Warehouse+Name=&Spend+Date=after+2020%2F02%2F01

PR #868, opened 2021-11-16
Add metered billing job

nickmisasi:
#### Summary
Adds a job that will run nightly to make POST requests to BLAPI for it to update usage of all subscriptions


dags/billing/nightly_metered_test.py
```diff
@@ -0,0 +1,56 @@
...
+)
+
+
+run_nightly_billing_cmd = f"""
+    {clone_and_setup_extraction_cmd} &&
+    python billing/run__metered_nightly_billing.py test
+"""
+
+if pod_defaults['is_delete_operator_pod'] == True:
+    pod_defaults['is_delete_operator_pod'] = False
```
neallred, line 45: Nit: Any reason not to set ` pod_defaults['is_delete_operator_pod'] = False` without checking? Can it be something other than a boolean?

nickmisasi: remnants of a debug. removed.

neallred (approved):
lgtm!

billing/run_metered_nightly_billing.py
```diff
@@ -0,0 +1,86 @@
...
+                    )
+                    break
+                except requests.exceptions.RequestException:
+                    message = f"Retrying: Error building invoice for subscription {sub['id']} with dates {start_date}-{end_date}"
+                    logging.error(message)
+
+            if resp and resp.status_code != requests.codes.ok:
+                message = f"Error building invoice for subscription {sub['id']} with dates {start_date}-{end_date}"
+                logging.error(message)
+                errors.append(message)
```
marianunez, line 77: Should we check this before we break in case it need to retry in some status code cases?

billing/run_metered_nightly_billing.py
```diff
@@ -0,0 +1,86 @@
...
+        for sub in conn.execute(query):
+            retries = 0
+            url = f"{blapi_url}/api/v1/customer/{sub['customer_id']}/subscriptions/{sub['id']}/usage"
+
+            resp = None
+            while retries < 10:
+                try:
+                    retries += 1
+                    resp = requests.post(
+                        url, json=payload, headers=header, timeout=(0.1, 10)
```
marianunez, line 67: Where do we get the usage ?

nickmisasi: This whole file is essentially just a cron job. It tells BLAPI to queue a usage reporting job and instantly returns. All of the calculation happens within BLAPI.

billing/run_metered_nightly_billing.py
nickmisasi: The only time this will not be 200 is if the customer or subscription doesn't exist: https://git.internal.mattermost.com/bizops/blapi/-/merge_requests/130/diffs#60a964aea4a6760d2f0fe80a7725ed9d569edd6c_327_339 unless there's some 

Both of which we can't really recover from at this point

billing/run_metered_nightly_billing.py
marianunez: Gotcha, for a second I thought I was in blapi repo.

marianunez (approved):

PR #872, opened 2021-11-23
Filter Cloud Starter, Professional, and Enterprise from onprem_subscriptions

nickmisasi:


marianunez (approved):
Interesting, wonder if there are any more queries where we would need to add the new skus

PR #873, opened 2021-11-24
removing the demandbase account field

nat-gunner:


nat-gunner:
Removed the demandbase account field from the coalesce statement in `hightouch.lead_update_account_link` since there's a demandbase issue that's causing leads to be associated with the wrong account.

PR #881, opened 2022-01-13
Update helpers.sql

sri-byte:
Added new filter to macro in helpers.sql to remove future dates from user_events_telemetry

### DIFF
--- a/transform/snowflake-dbt/macros/helpers.sql
+++ b/transform/snowflake-dbt/macros/helpers.sql
@@ -393,6 +393,7 @@ select get_sys_var({{ var_name }})
                                             'application_installed',
                                             'application_opened',
                                             'application_updated')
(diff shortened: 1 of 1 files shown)


sri-byte:
1) There are no side effects, as we are merging data from 7 different tables into one table, which is the user_events_telemetry table. 
The original data still resides in the 7 source tables. In the future when we will get to the source of the issue, we are seeing with future dates, this filter would no longer be required and subsequently removed.

2) The change has been fully tested locally on the database, and it works as expected.

jasonblais:
Great, thanks @sri-byte! 

@nickmisasi are you open to approving and merging the changes? Looks like I don't have permissions to (which is okay).

nickmisasi (approved):

PR #882, opened 2022-01-19
Breaking CTE's in license_server_fact to their own individual tables.

sri-byte:
The data in license_server_fact remains the same. The CTE's have their own individual tables with this commit and license_server_fact references those tables, due to which there is no impact on license_server_fact.

This improves readability as well as helps us find and debug any data issues, and improves data quality and improves run time.

### DIFF
--- a/transform/snowflake-dbt/models/blp/license_server_fact.sql
+++ b/transform/snowflake-dbt/models/blp/license_server_fact.sql
@@ -12,109 +12,14 @@
 
 {% endif %}
 
-with account_mapping as (
-  SELECT 
-      elm.account_sfid
-    , a.name as account_name
-    , elm.licenseid as license_id
-    , elm.opportunity_sfid
-    , elm.company
-    , elm.contact_sfid
-    , elm.edition
-  FROM (
-        SELECT
-            COALESCE(elm.account_sfid, lo.account_sfid)         AS account_sfid
-          , COALESCE(elm.opportunity_sfid, lo.opportunity_sfid) AS opportunity_sfid
-          , COALESCE(trim(elm.licenseid), trim(lo.licenseid))   AS licenseid
-          , COALESCE(trim(elm.company), trim(lo.company))       AS company
-          , COALESCE(trim(lo.contact_sfid), NULL)       AS contact_sfid
-          , lo.edition AS edition
-        FROM {{ ref('enterprise_license_mapping') }} elm
-        FULL OUTER JOIN {{ ref('license_overview') }} lo
-          ON trim(elm.licenseid) = trim(lo.licenseid)
-        GROUP BY 1, 2, 3, 4, 5, 6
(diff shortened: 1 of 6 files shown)


transform/snowflake-dbt/models/blp/licensed_servers.sql
```diff
@@ -0,0 +1,51 @@
...
+  , MIN(s.first_active_date) AS server_activation_date
+FROM {{ ref('licenses') }} l
+LEFT JOIN {{ ref('server_fact') }} s
+  ON l.server_id = s.server_id
+LEFT JOIN  {{ ref('account_mapping') }} am
+  ON l.license_id = am.license_id
+WHERE l.server_id IS NOT NULL
+AND l.license_id <> '16tfkttgktgdmb5m8xakqncx3c'
+AND l.license_id <> 'mdhgp9rsjtyfjeye1nhrtgqapo'
+AND l.license_id <> 'xmfoy61ru3r55kcosr9bpyw35h'
```
nickmisasi, line 39: I think you can remove these 2 conditions. I put them there and they're not needed. `16tfkttgktgdmb5m8xakqncx3c` is from before my time, so not sure what it's for. Maybe we can remove that too and see what happens?

nickmisasi (changes requested):
Great work @sri-byte! One thing to address

jasonblais:
@sri-byte Thank you! 

Propose adding these two notes into the PR description, as a best practice:

1. Do we see any risks or side effects this change may introduce?
2. Did we complete testing / verification of this change?

sri-byte:
@jasonblais 

1. The dbt materialization settings for the 3 new tables I set intially to 'Incremental', this will not work as we have the ID column which is being created as a combination of server_id and license_id, and in the data we have duplicates. This was identified and fixed by changing the setting to 'table'
2. The changes have been fully tested locally and in production.

nickmisasi (approved):

jasonblais:
@sri-byte thanks! Can you help include the above two notes in your future PR descriptions :+1:

transform/snowflake-dbt/models/blp/license_server_fact.sql
```diff
@@ -12,109 +12,14 @@
...
-        GROUP BY 1, 2, 3, 4, 5, 6
-      ) elm
-  LEFT JOIN {{ ref( 'account') }} a
-      ON elm.account_sfid = a.sfid
-  GROUP BY 1, 2, 3, 4, 5, 6, 7
-),
-
-licensed_servers as (
-SELECT
-    {{ dbt_utils.surrogate_key(['l.license_id', 'l.server_id']) }} as id
```
sri-byte, line 44: We don't need to add account_mapping since the unique ID is being generated in licensed_servers and nonactivated_licenses. @jasonblais

PR #883, opened 2022-01-28
Reverting last PR breaking license table

sri-byte:
In the last PR, I broke the license tables CTE's into their own individual tables. I also changed the license table to read these individual tables. The table has a 'unique' constraint on ID, and bad data trying to creep in is breaking the scheduled jobs.  We are still keeping the small tables, but reverting the license table to CTE's.

1) Impact of this is the dbt scheduled jobs will no longer fail. We will have to go back to see why we're getting bad data.
2) This fix is temporary until we figure out the source of the bad data.

### DIFF
--- a/transform/snowflake-dbt/models/blp/license_server_fact.sql
+++ b/transform/snowflake-dbt/models/blp/license_server_fact.sql
@@ -12,14 +12,75 @@
 
 {% endif %}
 
-with license_union as (
-  SELECT *
-  FROM {{ ref('licensed_servers') }}
-  
+with licensed_servers as (
+SELECT
+    {{ dbt_utils.surrogate_key(['l.license_id', 'l.server_id']) }} as id
+  , l.server_id
+  , l.license_id
+  , MAX(trim(COALESCE(am.company, l.company, s.company))) AS company
+  , MAX(COALESCE(am.edition, l.edition)) AS edition
+  , MAX(l.users)   AS users
+  , l.trial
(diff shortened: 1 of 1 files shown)


transform/snowflake-dbt/models/blp/license_server_fact.sql
```diff
@@ -12,14 +12,75 @@
 
 {% endif %}
 
-with license_union as (
-  SELECT *
-  FROM {{ ref('licensed_servers') }}
-  
+with licensed_servers as (
```
jasonblais, line 15: Do we need to add `account_mapping` to this file as well?

https://github.com/mattermost/mattermost-data-warehouse/pull/882/files#diff-0e46383f6e4ddebd5d363b6333e4d964e162a9417022aefa3d31799cbc73c560L15-L41

sri-byte:
```
SELECT
    {{ dbt_utils.surrogate_key(['l.license_id', 'l.server_id']) }} as id
```
We don't need account_mapping as the unique constraint, which is a combination of license_id and server_id, is created after account_mapping is created.

jasonblais: @sri-byte Moving your comment over here

> We don't need to add account_mapping since the unique ID is being generated in licensed_servers and nonactivated_licenses. @jasonblais

Got it, thanks for confirming it Sri 👍

jasonblais (approved):

PR #887, opened 2022-02-14
MM-41737 Adding community MM data from mm_telemetry_rc to performance_events

sri-byte:
1) Impact: `EVENTS.PERFORMANCE_EVENTS` is only looking at production data. Community Mattermost is sending data to `MM_TELEMETRY_RC`, this PR apends Community Mattermost data from `MM_TELEMETRY_RC ` to Production data. 
The source of the data which is `MM_TELEMETRY_RC` can be derived from `_dbt_source_relation` column, so there is a clear distinguish between `prod `and `rc`. There is no impact to production data, as we are only adding new data.
2) Testing: The changes have been tested fully in my local environment.

### DIFF
--- a/transform/snowflake-dbt/macros/helpers.sql
+++ b/transform/snowflake-dbt/macros/helpers.sql
@@ -58,7 +58,8 @@ select get_sys_var({{ var_name }})
 
 {% macro license_cleaning(destination_table) %}
     {% set query %}
-        delete from analytics.blp.license_server_fact using (SELECT license_id FROM analytics.blp.license_server_fact WHERE server_id IS NOT NULL GROUP BY 1) lsf
+        delete from analytics.blp.license_server_fact using (SELECT license_id FROM analytics.blp.license_server_fact 
+        WHERE server_id IS NOT NULL GROUP BY 1) lsf
         WHERE license_server_fact.license_id = lsf.license_id AND license_server_fact.server_id is null
     {% endset %}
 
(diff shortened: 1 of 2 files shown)


jasonblais:
> There is no impact to production data, as we are only adding new data.

@sri-byte If we are adding new data from the RC data source, would future production data be impacted?

sri-byte:
They each have their own source, we are combining the 2 sources, the query generated by this PR would have 2 CTE's one from Prod and one from RC (which is only prod at the moment). Also, from the RC one we are only pulling community.mattermost instance.

jasonblais:
> They each have their own source, we are combining the 2 sources, the query generated by this PR would have 2 CTE's one from Prod and one from RC (which is only prod at the moment). Also, from the RC one we are only pulling community.mattermost instance.

Sounds good. The RC key might have a lot of noisy test data, but if we are only adding community.mattermost instance, this should be fine.

jasonblais (approved):
Approving with the note below. 

> They each have their own source, we are combining the 2 sources, the query generated by this PR would have 2 CTE's one from Prod and one from RC (which is only prod at the moment). Also, from the RC one we are only pulling community.mattermost instance.

catalintomai:
@sri-byte - maybe for another PR - but to me MM_TELEMETRY_RC is quite general - maybe a scrub on naming would be good - defining a taxonomy across all Rudderstack et al. sources

transform/snowflake-dbt/macros/helpers.sql
```diff
@@ -305,7 +306,16 @@ select get_sys_var({{ var_name }})
                  FROM {{ this }} 
                  WHERE {{this}}.received_at <= CURRENT_TIMESTAMP
                  GROUP BY 1
-             ), 
+             ),
+        {%+ elif this.table == 'performance_events' %}
```
catalintomai, line 310: could we define a static var for 'performance_events' - might be useful for testing as well (have to rename the table to smth like `performance_events_test` in only one place)

PR #888, opened 2022-02-16
MM-41635 Add Retention 1 day user counts to server_fact table

sri-byte:
1) Impact: We are trying to understand how many users are active after 24 hours of their first appearance on a Mattermost Instance. To achieve this I am adding a new column which is similar to the `retention_28days_users` count with a minor change in logic. This new field will be used as a metric in the Growth Scorecard dashboard.
2) Testing: Changes have been tested locally and the column data was also validated and works as expected.


jasonblais:
Thanks @sri-byte!  Is this measuring at the user level or at the workspace/server-level? Eventually we'll want to measure both, but measuring it at the workspace/server-level is a priority.

sri-byte:
@jasonblais This is measuring at Server level.

jasonblais (approved):

transform/snowflake-dbt/models/mattermost/hourly/server_fact.sql
```diff
@@ -397,6 +399,7 @@ WITH sdd AS (
         , MAX(server_details.last_active_user_date) AS last_active_user_date
         , MAX(lsd.retention_28day_flag) AS retention_28day_flag
         , MAX(lsd.retention_28day_users) AS retention_28day_users
+        , MAX(lsd.retention_1day_users) AS retention_1day_users
```
catalintomai, line 402: @sri-byte  - nit: reorder in increasing order (1 day, 28 day, etc.)

catalintomai (approved):

nickmisasi (approved):

PR #892, opened 2022-03-07
Create user_retention table for New User Retention Metrics

sri-byte:
Impact: Creating new table user_retention for Growth Scoreboard. This is similar to the user_28day_retention table, the limitation of user_28day_retention is that it only serves Day 28 Retention metrics. My table can be modified/updated to add new user retention metrics. Currently serves Day 1, 7, 14 and 28 user retention metrics.
Testing: The data has been fully validated in Snowflake and the visualizations were created for the Growth Scoreboard.

### DIFF
--- a/transform/snowflake-dbt/models/mattermost/hourly/user_retention.sql
+++ b/transform/snowflake-dbt/models/mattermost/hourly/user_retention.sql
@@ -1,0 +1,52 @@
+{{config({
+    "materialized": "incremental",
+    "schema": "mattermost",
+    "tags":"union",
+    "snowflake_warehouse": "transform_m",
+    "unique_key":"id"
+  })
+}}
+
+WITH first_active AS (
+    SELECT
+        uet.user_actual_id
+      , sf.server_id
+      , {{ dbt_utils.surrogate_key(['uet.user_actual_id', 'sf.server_id'])}} AS id
+      , sf.installation_id
+      , sf.first_server_edition
+      , MIN(uet.timestamp) AS first_active_timestamp
+    FROM {{ ref('user_events_telemetry') }} uet
+         JOIN {{ ref('server_fact') }}  sf
+              ON COALESCE(uet.user_id, IFF(LENGTH(uet.context_server) < 26, NULL, uet.context_server),
+                          IFF(LENGTH(uet.context_traits_userid) < 26, NULL, uet.context_traits_userid),
+                          IFF(LENGTH(uet.context_server) < 26, NULL, uet.context_server)) = sf.server_id
+            AND uet.timestamp::date >= sf.first_active_date::date
+            AND sf.first_active_date::date >= '2020-02-01'
+    {% if is_incremental() %}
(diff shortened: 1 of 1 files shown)


transform/snowflake-dbt/models/mattermost/hourly/user_retention.sql
```diff
@@ -0,0 +1,52 @@
...
+      , sf.installation_id
+      , sf.first_server_edition
+      , MIN(uet.timestamp) AS first_active_timestamp
+    FROM {{ ref('user_events_telemetry') }} uet
+         JOIN {{ ref('server_fact') }}  sf
+              ON COALESCE(uet.user_id, IFF(LENGTH(uet.context_server) < 26, NULL, uet.context_server),
+                          IFF(LENGTH(uet.context_traits_userid) < 26, NULL, uet.context_traits_userid),
+                          IFF(LENGTH(uet.context_server) < 26, NULL, uet.context_server)) = sf.server_id
+            AND uet.timestamp::date >= sf.first_active_date::date
+            AND sf.first_active_date::date >= '2020-02-01'
```
catalintomai, line 24: should we move this out as a var/const before, to be referenced?

transform/snowflake-dbt/models/mattermost/hourly/user_retention.sql
```diff
@@ -0,0 +1,52 @@
...
+    {% if is_incremental() %}
+    WHERE uet.timestamp < CURRENT_TIMESTAMP
+      AND {{ dbt_utils.surrogate_key(['uet.user_actual_id', 'sf.server_id'])}} NOT IN (SELECT id FROM {{this}} group by 1)
+      AND uet.user_actual_id IS NOT NULL
```
catalintomai, line 28: can we factor this out?

transform/snowflake-dbt/models/mattermost/hourly/user_retention.sql
```diff
@@ -0,0 +1,52 @@
...
+    {% else %}
+    WHERE uet.timestamp < CURRENT_TIMESTAMP
+      AND uet.user_actual_id IS NOT NULL
+    {% endif %}
```
catalintomai, line 32: wondering if this can be simplified or not?

catalintomai (approved):

transform/snowflake-dbt/models/mattermost/hourly/user_retention.sql
sri-byte: I don't know if that is possible in dbt/sql. we can do it using Stored Procedures, but that would make things really complex

transform/snowflake-dbt/models/mattermost/hourly/user_retention.sql
sri-byte: this gets compiled and these statements don't go through, I'll see if we can simplify this further

nickmisasi (approved):

PR #894, opened 2022-03-14
Adding Distinct to remove duplicates from the user_retention table

sri-byte:
Impact: Without distinct there are a lot of duplicate rows in the table, which is added processing for Looker.

Testing: The table has been cleaned up using the new code.

### DIFF
--- a/transform/snowflake-dbt/models/mattermost/hourly/user_retention.sql
+++ b/transform/snowflake-dbt/models/mattermost/hourly/user_retention.sql
@@ -32,7 +32,7 @@ WITH first_active AS (
     {% endif %}
     GROUP BY 1, 2, 3, 4, 5
 )
-SELECT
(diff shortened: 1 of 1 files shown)


jasonblais:
Is it going to impact the actual metrics in the scorecard -- e.g. by removing duplicate rows, would we expect higher or lower user retention numbers?

sri-byte:
No actual metrics would be impacted, I reviewed the data, we are having the same rows multiple times due to the join condition on the fact and event table. The size of the table reduces improving query performance for Looker with this change.

sri-byte:
The first active date is also a timestamp, so the duplicates are at the timestamp level.

jasonblais (approved):
Approving per note that there is no impact on actual metrics & reporting, only on query performance.

PR #905, opened 2022-04-21
MM - 43454 Fixing server_fact and adding a server to excludable_server_seed

sri-byte:
Impact: Server_fact table which uses server_daily_details as source, is dropping new servers. I observed this when I did a full refresh of the server_fact table and all the plots got updated with this new server data. Adding a not exists clause fixes this issue and server_fact wouldn't drop any rows. Also adding an exclusion server to excludable server seed file.

Testing: Changes have been tested locally and this PR works as expected.

image
image


jasonblais (approved):
Thanks @sri-byte!

jasonblais:
PS: I wonder if we would begin to get a second review from the data engineer on PRs in mattermost-data-warehouse repo, as a way to familiarize with data warehouse changes. 0/5

PR #906, opened 2022-04-21
Update to server_fact fix from the last PR #905

sri-byte:
Impact: There was an unforeseen issue which made its way in, in the last PR merge. The missing new servers from the 24 hour telemetry job got fixed and the servers are getting added, but the job is setting the first active date to the current date to a lot of servers, which is causing a spike in the charts. 
I will have to re-think my solution to fix the missing new servers.

Testing: This change reverts the job to the way it was before.

Update: Added the missing servers condition to the WHERE clause and reverted to the original HAVING condition.

### DIFF
--- a/transform/snowflake-dbt/models/mattermost/hourly/server_fact.sql
+++ b/transform/snowflake-dbt/models/mattermost/hourly/server_fact.sql
@@ -30,7 +30,7 @@ WITH sdd AS (
             , MAX(CASE WHEN in_mm2_server THEN timestamp ELSE NULL END)                           AS last_mm2_telemetry_date
             FROM {{ ref('server_daily_details') }} SDD
             WHERE timestamp <= CURRENT_TIMESTAMP 
-            OR NOT EXISTS (SELECT 1 FROM {{this}} SF WHERE SF.SERVER_ID = SDD.SERVER_ID)
+            OR NOT EXISTS (SELECT 1 FROM {{this}} SF WHERE SF.SERVER_ID = SDD.SERVER_ID AND SDD.timestamp <= CURRENT_TIMESTAMP)
             GROUP BY 1
             {% if is_incremental() %}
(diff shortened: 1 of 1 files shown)


transform/snowflake-dbt/models/mattermost/hourly/server_fact.sql
```diff
@@ -31,9 +31,7 @@ WITH sdd AS (
             FROM {{ ref('server_daily_details') }} SDD
             WHERE timestamp <= CURRENT_TIMESTAMP
             {% if is_incremental() %}
-            AND ((CASE WHEN in_security OR in_mm2_server THEN timestamp ELSE NULL END) > (SELECT MAX(last_active_date) - INTERVAL '2 HOURS' FROM {{this}})
-            OR NOT EXISTS (select 1 FROM {{this}} SF WHERE SF.SERVER_ID = SDD.SERVER_ID)
-            )
+            AND (CASE WHEN in_security OR in_mm2_server THEN timestamp ELSE NULL END) > (SELECT MAX(last_active_date) - INTERVAL '2 HOURS' FROM {{this}})
```
jasonblais, line 34:
```suggestion
            HAVING MAX(CASE WHEN in_security OR in_mm2_server THEN timestamp ELSE NULL END) > (SELECT MAX(last_active_date) - INTERVAL '2 HOURS' FROM {{this}})
```

jasonblais: ^this is what the query used to be

jasonblais (changes requested):

transform/snowflake-dbt/models/mattermost/hourly/server_fact.sql
sri-byte: you are correct about that, I'm testing the impact of adding the other condition to add the missing servers here.

jasonblais (approved):

PR #912, opened 2022-05-03
Removing DLL function to avoid DBT error

RoyI99:


RoyI99:
Hi @sri-byte I deleted the dll create table function so it does not cause an error on DBT.

sri-byte (approved):

RoyI99:
@sri-byte please merge this commit.  thanks!

PR #923, opened 2022-05-11
Add ldap and custom user groups to server_daily_details_ext

sri-byte:
Testing: Minor code changes to previous commit after running more data tests.


mkraft (commented):
What do the numbers `600`, `581`, and `28` represent?

sri-byte:
the code gets compiled using this macro, https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/groupby.sql, where the number n=600, 581 and 28 represents the number of fields used to group by, when compiling the code.

mkraft (approved):

PR #925, opened 2022-05-18
MM-42451-Freemium-cloud-subscriptions-flow 

munish7771:
#### Summary
Added Blapi and Hightouch models for cloud freemium data flow.


transform/snowflake-dbt/models/hightouch/freemium/freemium_account_update.sql
```diff
@@ -0,0 +1,21 @@
+{{config({
+    "schema": "hightouch",
```
catalintomai, line 2: @munish7771 , what is the difference from  `freemium_account.sql'? Also, why is `freemium_account_cancelled.sql` empty?

munish7771: instead of upsert, we do updates and inserts in different syncs. Accounts that already exist in salesforce will be updated.

munish7771: cancelled subscriptions is a part of another task. Accidently added file in this pr, removing it.

catalintomai: > instead of upsert, we do updates and inserts in different syncs. Accounts that already exist in salesforce will be updated.

didn't see any difference code-wise, therefore the q.

munish7771: theres a diff in filter, 
` WHERE account.id IS NOT NULL AND account_domain_mapping.accountid IS NOT NULL` for account_update

sri-byte (approved):

PR #931, opened 2022-05-26
Bugfix to remove duplicates associated with multiple opportunity owner values

RoyI99:
Removed opportunity owner field as it causes duplicates with multiple values


RoyI99:
@sri-byte kindly approve the pr above.

sri-byte (approved):

PR #932, opened 2022-05-30
MM - 43824 Adding total boards, views and cards to server_fact

sri-byte:
Impact: Adding total boards, views and cards to server_fact from Focalboard blocks, for Freemium Workspace limits.

Testing: Changes have been fully tested and columns were developed in Looker Viz as well.

### DIFF
--- a/transform/snowflake-dbt/models/mattermost/hourly/server_fact.sql
+++ b/transform/snowflake-dbt/models/mattermost/hourly/server_fact.sql
@@ -76,7 +76,17 @@ WITH sdd AS (
       where activity.user_id in (SELECT server_id from sdd GROUP BY 1)
       group by 1
     ),
-
+    max_blocks_time AS (
+      select  
+          fs.server_id
+        , fb.user_id
+        , max(fb.timestamp)       as max_time
+        , max(fb.timestamp::date) as max_date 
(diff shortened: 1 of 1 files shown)


transform/snowflake-dbt/models/mattermost/hourly/server_fact.sql
```diff
@@ -169,9 +193,14 @@ WITH sdd AS (
         , max(COALESCE(r.outgoing_webhooks, s.outgoing_webhooks)) as outgoing_webhooks
         , max(COALESCE(r.max_time, s.max_time)) AS max_timestamp
         , MAX(COALESCE(r.context_ip, r.context_request_ip)) AS last_ip_address
+        , MAX(board) AS boards
+        , MAX(_view) AS boards_views
```
jasonblais, line 197: Why is this view with an underscore `_view`?

sri-byte: The table originally has an underscore. No idea why its like that, I can update it, but if its being used downstream or anyone's code across MM, it will break.

jasonblais: I see, yea we should look at making a change but not a priority for this PR

PR #933, opened 2022-06-02
Modifying Arr Roll Forward to Remove Opportunity Owner  and Add LTV Classifications

RoyI99:
As discussed with Srikar removing opportunity owner which did not appear as committed in the previous PR.  Also adding LTV classifiers for Finance.


RoyI99 (commented):
Tested and ran on snowflake

sri-byte (approved):

PR #942, opened 2022-06-15
Updating community in excludable servers

sri-byte:
Impact: Updating community in excludable servers so we can add only community along with other prod servers in various Analyses.


agarciamontoro (approved):
:tada:

PR #944, opened 2022-06-16
Updated models for campaignmember status

munish7771:
Fixed stripe edition error and campaignmember status.

#### Summary
TL;DR
1. Updated campaignmember status with accepted values.
2. Updated the filter returning cloud starter leads.
3. Added lead_source and lead_source fields in sso lead model.

BREAKING
1. Campaignmember status had some values that were not accepted from salesforce, causing models to fail.
2. Stripe field "edition" was coming as null for some records, causing filter on it to return less number of rows.

#### Ticket Link
N/A

### DIFF
--- a/transform/snowflake-dbt/models/hightouch/freemium/cloud_starter/cs_sso_facts.sql
+++ b/transform/snowflake-dbt/models/hightouch/freemium/cloud_starter/cs_sso_facts.sql
@@ -40,6 +40,7 @@ with existing_leads as (
     subscriptions.created,
     subscriptions.trial_start,
     subscriptions.trial_end,
+    coalesce(subscriptions.edition, products.name) as edition,
     coalesce(campaignmember.dwh_external_id__c, UUID_STRING('78157189-82de-4f4d-9db3-88c601fbc22e', '7013p000001TxBuAAK' || customers.email)) AS campaignmember_external_id,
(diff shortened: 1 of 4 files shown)


transform/snowflake-dbt/models/hightouch/freemium/cloud_starter/cs_sso_facts.sql
```diff
@@ -53,17 +54,19 @@ with existing_leads as (
...
         left join existing_leads as lead on customers.email = lead.email and lead.row_num = 1
         left join existing_contacts as contact on customers.email = contact.email and contact.row_num = 1
         left join existing_members as campaignmember on customers.email = campaignmember.email and campaignmember.row_num = 1
-        where customers.email not in ( select email from {{ ref('cs_signup_campaign')}} )
-        and lower(subscriptions.edition) = 'cloud starter' and lead_sync_eligible 
+        where customers.email not in ( select email from {{ ref('cs_signup_campaign_dev')}} ) and lead_sync_eligible 
 )
-select * from sso_facts_pre
\ No newline at end of file
+select * from sso_facts_pre where lower(edition) = 'cloud starter'
```
catalintomai, line 72: @munish7771 - is it `subscriptions.edition` (like before) or only `edition`

munish7771: column is from same cte sso_facts_pre, reference is not needed. only edition.

nat-gunner (approved):
looks good to me!

PR #946, opened 2022-06-20
MM-42447 updated sso leads for cloud starter campaign

munish7771:
#### Summary
TL;DR

1. Added sso leads from raw source 'identifies'
2. Added sso_provider and is_sso columns to model.
3. Fixed company_name[missing] and first_name[data value too large errors], since these errors were due to previously used sso models which is now depreciated.
4. Removed the cloud starter filter which was filtering out incomplete signups.

Previously we were getting sso leads from stripe, now we are getting it from portal and able to track the signup process for them.


catalintomai (approved):

sri-byte (approved):

sri-byte:
Nice work

PR #949, opened 2022-06-27
Update lead status

munish7771:
#### Summary
Bug - Lead statusis coming as "account verified" or "workspace created". "Email verified" was getting skipped due to incorrect sequence in case statement. 

Fix - Updated the case statement.


sri-byte (approved):

nat-gunner (approved):
thumbs up from me!

PR #952, opened 2022-07-06
MM-42451 professional subscriptions and cancellation workflow

munish7771:
#### Summary
Updates - 
1. Created a new view containing customers with cloud paid subscriptions which sources subscriptions from stripe. Updated the hightouch models accordingly.
2. Updated hightouch models with correct account_external_id and opportunity_external_ids.
3. Added columns order_type and subscription_id for professional subscriptions.
4. Added models for cancelled subscriptions. (only cloud professional)


nat-gunner (approved):
everything looks good as far as I can tell!

PR #956, opened 2022-07-11
Update version_release_dates.csv

amyblais:
- 6.4 and 6.5 are out of support, v6.6 is going out of support in a few days (July 15).


rmonga71 (approved):
Thanks Amy - looks good to me

sri-byte (approved):

PR #958, opened 2022-07-13
MM - 45619 Adding Day 0 flag to server_fact and user_retention

sri-byte:
Impact; Adding Day 0 flag to server_fact and user_retention


dontoisme (commented):
@sri-byte is there anything here about removing the `performance` category?

dontoisme (approved):

PR #960, opened 2022-07-15
Don't keep BLAPI import job pods after successful execution

nickmisasi:
#### Summary
This was set for debugging and not unset when debugged. We don't need to keep these pods around.


nickmisasi:
@sri-byte @munish7771 if you are the final approver, please merge it!

sri-byte (approved):

PR #964, opened 2022-07-19
Updating ARR_REPORT to include invoice calculation subquery

RoyI99:
Updating ARR_REPORT to cover request by T Quock to add calculated invoice billing for the first year.  Executed by adding a subquery called inv


RoyI99 (commented):
tested to work on snowflake

sri-byte (approved):

PR #965, opened 2022-07-19
MM - 45847 Cleaning up user_events_telemetry table

sri-byte:
Impact: Dropping unused columns from user_events_telemetry table, these columns are created by rudderstack and have no use for analytics. They only add a lot of columns to the table and add complexity.
Testing: Changes have been tested locally.


jasonblais (approved):
Thank you Sri :+1:

PR #968, opened 2022-07-25
Fixing Boards join in server fact

sri-byte:
Impact: The boards join in server_fact was using timestamp, which wasn't returning any rows. Fixed the join to use date.


munish7771 (approved):
lgtm

PR #973, opened 2022-08-08
Appended Null Salesforce License Dates with Stripe Data

RoyI99:
Portal data on license issued for self-service is not feeding Salesforce.
Appended ARR Transactions for license dates from Stripe Invoice Activity (subscription table)


RoyI99:
@sri-byte please review and approve

sri-byte (approved):

PR #974, opened 2022-08-09
Importing secrets for nps 

munish7771:
#### Summary
Update - Secrets injected into pod need to be imported from env.

### DIFF
--- a/utils/post_nps_data.py
+++ b/utils/post_nps_data.py
@@ -8,13 +8,13 @@ def format_row(row):
     return tuple(row)
 
 conn = snowflake.connector.connect(
-    user = SNOWFLAKE_USER,
-    password = SNOWFLAKE_PASSWORD,
-    account = SNOWFLAKE_ACCOUNT,
-    warehouse = SNOWFLAKE_TRANSFORM_WAREHOUSE,
-    database = SNOWFLAKE_TRANSFORM_DATABASE,
-    schema = SNOWFLAKE_TRANSFORM_SCHEMA
-            )
+    user = os.getenv("SNOWFLAKE_USER"),
+    password = os.getenv("SNOWFLAKE_PASSWORD"),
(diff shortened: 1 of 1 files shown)


sri-byte (approved):

utils/post_nps_data.py
```diff
@@ -8,13 +8,13 @@ def format_row(row):
...
-    warehouse = SNOWFLAKE_TRANSFORM_WAREHOUSE,
-    database = SNOWFLAKE_TRANSFORM_DATABASE,
-    schema = SNOWFLAKE_TRANSFORM_SCHEMA
-            )
+    user = os.getenv("SNOWFLAKE_USER"),
+    password = os.getenv("SNOWFLAKE_PASSWORD"),
+    account = os.getenv("SNOWFLAKE_ACCOUNT"),
+    warehouse = os.getenv("SNOWFLAKE_TRANSFORM_WAREHOUSE"),
+    database = os.getenv("SNOWFLAKE_TRANSFORM_DATABASE"),
+    schema = os.getenv("SNOWFLAKE_TRANSFORM_SCHEMA")
```
nickmisasi, line 16: You should import these from: https://github.com/mattermost/mattermost-data-warehouse/blob/7bd8cc8f1690027a50d5b9f63fa8abc4c548f78d/dags/kube_secrets.py#L61

Don't access the env directly.

nickmisasi (changes requested):

utils/post_nps_data.py
nickmisasi: Like this: https://github.com/mattermost/mattermost-data-warehouse/blob/7bd8cc8f1690027a50d5b9f63fa8abc4c548f78d/dags/billing/nightly.py#L14

munish7771: Unable to access dags from inside the pod.

nickmisasi (approved):

PR #983, opened 2022-08-18
Updating Logic for Contract Count, New ARR and Active Count

RoyI99:
Updating logic as further use cases for parent child id are discovered


RoyI99 (commented):
Removed account_source filed from arr_rollforward

PR #987, opened 2022-08-26
Simplify week to snowflake convention and create renew field that excludes expansion

RoyI99:
Updating arr_transaction and reporting to the logic on header.  Tested on snowflake


RoyI99 (commented):
tested on snowflake

sri-byte (approved):

PR #1002, opened 2022-09-30
MM - 47174 Fix clearbit cloud airflow dag failure

sri-byte:
Impact: Fix clearbit cloud airflow dag failure.
Testing: Tested end to end on local machine.

Fix requires adding the columns to the backend table which don't exist in the Clearbit enrichment data.


jasonblais:
@sri-byte what are your thoughts on having Henry to review, so he learns about the code changes as part of the review?

PR #1012, opened 2022-10-12
Fix onprem opportunitylineitem

munish7771:
#### Summary

Issue - Renewals are getting synced with the new subscription model. Since new subscription model runs before renew, it leads to incorrect data.

Updates - Added a filter


sri-byte (approved):

ifoukarakis (commented):
LGTM

PR #1014, opened 2022-10-13
MM-47325: replace dbt-run with cloud dbt job trigger

ifoukarakis:
#### Summary

- [x] Replace nightly dbt run using `KubernetesOperator` with triggering run at DBT cloud.
- [x] Setup basic test environment & dependencies using [Poetry](https://python-poetry.org/).
- [x] Introduce tests for code that handles DBT cloud execution.
- [x] Update `README.md` with instructions on how to run tests.

Please ignore the size of the PR. Most of it is due to the `poetry.lock`. This file is required in order to pin dependency versions. It is generated while dependencies are executed, but unfortunately is rather big.

Note that the dag might fail because it also triggers dbt-seed. This task is not needed, but will be commented out in a separate PR.

#### Ticket Link

Fixes [MM-47325](https://mattermost.atlassian.net/browse/MM-47325).


ifoukarakis:
Closing in order to break into smaller PRs.

PR #1015, opened 2022-10-13
MM-47325: manage dependencies using poetry

ifoukarakis:
#### Summary

- [x] Setup dependencies using [Poetry](https://python-poetry.org/). Poetry is a modern alternative to setuptools or plain `requirements.txt` file.
- [x] Add dependencies from existing docker image.
- [x] Update `README.md` with instructions on how to install dependencies.

Please treat this PR as a suggestion. Currently there's no dependency management. All dependencies are pre-installed in a hard-coded docker image. 

Also please ignore the size of the PR. Most of it is due to the `poetry.lock`. This file is required in order to pin dependency versions. It is generated while dependencies are executed, but unfortunately is rather big.

#### Ticket Link

Used to create dev environment for [MM-47325](https://mattermost.atlassian.net/browse/MM-47325).

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,9 @@ data to other tools.
   * [Setup](#setup)
   * [Starting a bash shell with DBT](#starting-a-bash-shell-with-dbt)
   * [Building docs and serving them locally](#building-docs-and-serving-them-locally)
+* [Developing](#developing)
+  * [Requirements](#requirements)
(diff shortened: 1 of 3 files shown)


nickmisasi (approved):
+1 on dependency management!

catalintomai (approved):

munish7771 (approved):
I like poetry!

PR #1016, opened 2022-10-13
MM-47325: add unit tests for existing utility

ifoukarakis:
#### Summary

Introduces tests for `run_dbt_cloud_job.py`.

The tests were written with the intent of not changing or improving the code in any manner. This way future changes in the code can be achieved with a bit more confidence. Unfortunately this has the side-effect that tests take a bit more to run, as timeout wait interval is hard coded. This should be fixed in a follow up PR.

#### Ticket Link

Tests code that triggers dbt cloud job, required for [MM-47325](https://mattermost.atlassian.net/browse/MM-47325).


nickmisasi (approved):
LGTM! Is the plan to run these in CircleCI at some point?

ifoukarakis:
> LGTM! Is the plan to run these in CircleCI at some point?

Yes, definitely be part of CI/CD! We are figuring the details on which tool to use.

tests/utils/test_run_dbt_cloud_job.py
```diff
@@ -0,0 +1,131 @@
...
+
+import pytest
+from responses import Response
+
+from utils.run_dbt_cloud_job import trigger_dbt_run, poll_dbt_run
+
+
+def test_should_trigger_run(responses):
+    # GIVEN: DBT cloud ready to accept job
+    rsp = Response(method="POST", url="https://cloud.getdbt.com/api/v2/accounts/1001/jobs/101/run/",
```
catalintomai, line 12: @ifoukarakis , can we store this url in a static var of some sort - it's being reused everywhere.

tests/utils/test_run_dbt_cloud_job.py
```diff
@@ -0,0 +1,131 @@
...
+    assert run_id == 10000
+
+
+@pytest.mark.parametrize("status,content", [(400, 'job.400.json'), (404, 'job.404.json')])
+def test_should_handle_failure(responses, status, content):
+    # GIVEN: DBT cloud returns a non-200 response
+    rsp = Response(method="POST", url="https://cloud.getdbt.com/api/v2/accounts/1001/jobs/101/run/",
+                   status=status, headers={"Authorization": f"Token test-dbt-key", "Content-Type": "application/json"},
+                   body=with_response_body(content))
+    responses.add(rsp)
```
catalintomai, line 34: @ifoukarakis , can we factor out this block into a method and pass the body content as a param?

ifoukarakis: That's a really good idea. I used the [factories as fixtures](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html?highlight=factories#factories-as-fixtures) pattern.

catalintomai (approved):

munish7771 (approved):

PR #1029, opened 2022-10-26
POC Alerting and monitoring for stitch

munish7771:
Consists of adding plugins
1. mattermost webhook operator -> posts messages to mattermost channel.
2. stitch sensor -> looks for failing integrations in stitch and alerts to mattermost channel using mattermost operator.

#### Summary

To run - 
1. Add mattermost web hook connection, web hook url in host param.
2. Add stitch api token, stitch extractions url, stitch loads url, stitch_integrations (a dict of integration_id, integration_name),   in airflow variables.

TODO -
1. add unit tests

sample mattermost alert message link -> https://community.mattermost.com/private-core/pl/1j8spdwrnb89bnebj95s8ztuxe

### DIFF
--- a/plugins/operators/mattermost_operator.py
+++ b/plugins/operators/mattermost_operator.py
@@ -1,0 +1,30 @@
+from airflow.utils.decorators import apply_defaults
+import requests
+
+class MattermostWebhookOperator():
+
+    @apply_defaults
+    def __init__(self,
+                 http_conn_url=None,
+                 message=None,
+                 channel=None,
+                 username=None,
+                 **kwargs) -> None:
+        self.http_conn_url = http_conn_url
+        self.message = message
+        self.channel = channel
+        self.username = username
+
+    def execute(self, context): 
+        print("executing post in mattermost operator- ", self.http_conn_url)
+        payload='{"text": "%s", "channel": "%s"}' % (self.message, self.channel)
+        response = requests.post(
+            self.http_conn_url, data=payload.encode('utf-8'),
+            headers={'Content-Type': 'application/json'}
+        )
+        if response.status_code == 200: return True
+        return False
+
+    def printme(self, context):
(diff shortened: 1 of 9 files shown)


plugins/operators/mattermost_operator.py
```diff
@@ -0,0 +1,30 @@
+from airflow.utils.decorators import apply_defaults
+import requests
+
+class MattermostWebhookOperator():
```
ifoukarakis, line 4: According to the [docs](https://airflow.apache.org/docs/apache-airflow/1.10.15/howto/custom-operator.html) operators must be descendands of `BaseOperator`. There's two options here:
- `BaseOperator` is the parent class. This is the simplest approach.
-`SimpleHTTPOperator` is the parent class. This looks like a better approach, as all HTTP handling can be deferred to `SimpleHTTPOperator` methods. This internally uses `HttpHook` which handles HTTP failures, retries etc.

plugins/operators/mattermost_operator.py
```diff
@@ -0,0 +1,30 @@
...
+                 channel=None,
+                 username=None,
+                 **kwargs) -> None:
+        self.http_conn_url = http_conn_url
+        self.message = message
+        self.channel = channel
+        self.username = username
+
+    def execute(self, context): 
+        print("executing post in mattermost operator- ", self.http_conn_url)
```
ifoukarakis, line 19:
```suggestion
        self.log.info("executing post in mattermost operator")
```

A better alternative for `print` is airflow's native logging. In general it's better to rely on python's logging rather than print function. It 

`http_conn_url` contains the webhook URL which can be considered as sensitive information. Logging/printing sensitive information must be avoided.

ifoukarakis (commented):
Added a couple of comments regarding the design of the operators.

munish7771:
closing this PR as its no longer needed.

PR #1030, opened 2022-10-26
MM-47918: test diagnostics job

ifoukarakis:
#### Summary

Tests for diagnostics job. This job is copying data from S3 to snowflake. The actual code is in `stage_import.py`. `diagnostics_job.py` is just an entry point, so it was omitted for now.


munish7771 (approved):
lgtm!

PR #1032, opened 2022-10-27
MM-47920: test licenses job

ifoukarakis:
#### Summary

Test job that imports license data from S3. The job is nothing more than a trigger to Snowflake to copy data from an S3 stage. Copying stage has already been tested in #1030, so all this PR needs to do is that the expected parameters are passed to the trigger.

### DIFF
--- a/tests/extract/s3_extract/test_stage_import.py
+++ b/tests/extract/s3_extract/test_stage_import.py
@@ -1,7 +1,7 @@
 import pytest
 from mock import call
 from extract.s3_extract.stage_import import extract_from_stage, diagnostics_import, get_diagnostics_pattern, \
-    push_proxy_import, get_path, get_push_proxy_pattern
+    push_proxy_import, get_path, get_push_proxy_pattern, licenses_import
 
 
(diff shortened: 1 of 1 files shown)


tests/extract/s3_extract/test_stage_import.py
```diff
@@ -1,6 +1,6 @@
 import pytest
 from mock import call
-from extract.s3_extract.stage_import import extract_from_stage, diagnostics_import, get_diagnostics_pattern
+from extract.s3_extract.stage_import import extract_from_stage, diagnostics_import, get_diagnostics_pattern, licenses_import
```
catalintomai, line 3: @ifoukarakis , could we remove `diagnostics_import`?

catalintomai (approved):

tests/extract/s3_extract/test_stage_import.py
ifoukarakis: Can you please specify what you mean with remove? If you're talking about the DAG, it is currently used, so most likely is needed. If you're talking about separating these tests to two files, let's do any refactoring in a separate PR.

PR #1044, opened 2022-11-09
Fix dbt hourly job failure

sri-byte:
Impact: This fixes the hourly dbt job failure due to incorrect field data type calculation from source table. The value is too large to be used as a numeric.

Testing: Changes have been fully tested and verified locally.

image


ifoukarakis (approved):
LGTM. Only a couple of questions/comments.

- Did the type of `ticket_field`_id` changed somehow?
- I'm thinking that this repeatable max/case can be converted to Jinja. Just a suggestion for the future.

munish7771 (approved):

sri-byte:
The type of the field hasn't changed, in the source table the field is VARCHAR, but we're doing an int comparison in our model which is incorrect.

PR #1049, opened 2022-11-11
MM-47583: run tests for airflow code

ifoukarakis:
#### Summary

- [x] Run tests on airflow related code (dags and plugins) on PR and on merge on master.
- [x] Skip running python tests on airflow changes.
- [x] Move requirements for airflow under `build/` to be consistent.
- [x] Use same python version as the one deployed at production airflow (`3.6.12`).
- [x] Update dependencies to work with airflow's python version.
- [x] Add related tasks to Makefile.

### DIFF
--- a/.github/workflows/ci-python.yml
+++ b/.github/workflows/ci-python.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - master
     paths:
-      - '.github/**'
+      - '.github/workflows/ci-python.yml'
       - 'extract/**'
       - 'tests/**'
       - 'utils/**'
@@ -15,7 +15,7 @@ on:
     branches:
       - master
     paths:
-      - '.github/**'
+      - '.github/workflows/ci-python.yml'
       - 'extract/**'
       - 'tests/**'
       - 'utils/**'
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-python.yml
```diff
@@ -15,7 +15,7 @@ on:
     branches:
       - master
     paths:
-      - '.github/**'
+      - '.github/workflows/ci-python.yml'
       - 'extract/**'
       - 'tests/**'
```
munish7771, line 20: I have added airflow_utils tests under tests/dags/*.

So for airflow, each test will be inside specific directory. like dags/tests/* etc and for rest its under tests/* ?.

munish7771 (approved):
Looks good !

.github/workflows/ci-python.yml
ifoukarakis: I checked the tests for plugins and noticed they are under `plugins/tests`. This is a good approach, as it helps separate the test code for python scripts from airflow. Having said that, I'd expect all tests for dags to be under `dags`. Pytest's documentation also discusses [different layouts for tests](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#choosing-a-test-layout-import-rules).

PR #1052, opened 2022-11-14
MM-48217: lint python code

ifoukarakis:
#### Summary

Introduces linting for python code. The following checks are performed on push:
- [isort](https://github.com/PyCQA/isort) to sort imports alphabetically, and automatically separated into sections and by type.
- [black](https://github.com/psf/black) for checking code formatting.
- [flake8](https://flake8.pycqa.org/en/latest/) for style guide enforcement. A set of additional plugins are added for `flake8`
  - [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear) for finding likely bugs.
  - [flake8-comprehensions](https://github.com/adamchainz/flake8-comprehensions) for writing better dict/list/set comprehensions.
  - [flake8-simplify](https://github.com/MartinThoma/flake8-simplify) to indicate parts of the code that can be simplified.
  - [flake8-eradicate](https://github.com/wemake-services/flake8-eradicate) that checks for commented out code so that it can be removed.

> Note that this is just the setup. If this PR is merged, most of the checks will fail as a lot of code linting failures exist.

### DIFF
--- a/tests/utils/test_run_dbt_cloud_job.py
+++ b/tests/utils/test_run_dbt_cloud_job.py
@@ -2,7 +2,7 @@ import json
 
 import pytest
 
-from utils.run_dbt_cloud_job import trigger_dbt_run, poll_dbt_run
+from utils.run_dbt_cloud_job import poll_dbt_run, trigger_dbt_run
 
 EXPECTED_RUN_TRIGGER_URL = "https://cloud.getdbt.com/api/v2/accounts/1001/jobs/101/run/"
 EXPECTED_RUN_POLL_URL = "https://cloud.getdbt.com/api/v2/accounts/1001/runs/42/"
--- a/.github/workflows/ci-lint.yml
+++ b/.github/workflows/ci-lint.yml
@@ -1,7 +1,7 @@
 name: Python Linter
 
 on:
-  push:
+  pull_request:
     paths:
       - '.github/**'
       - 'dags/**'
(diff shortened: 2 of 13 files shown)


tests/utils/test_run_dbt_cloud_job.py
```diff
@@ -2,7 +2,7 @@
 
 import pytest
 
-from utils.run_dbt_cloud_job import trigger_dbt_run, poll_dbt_run
+from utils.run_dbt_cloud_job import poll_dbt_run, trigger_dbt_run
```
munish7771, line 5: should this change be included in this PR?

munish7771:
Are the changes in files test* because of linter?

tests/utils/test_run_dbt_cloud_job.py
ifoukarakis: > Are the changes in files test* because of linter?

Exactly. I included some examples of changes that might be required if we're to adapt this. If the fixes suggested by the linters are not added, then builds will fail. I can think of two ways to go forward:
- Add linters and fixes in a single PR. This will solve the problem, but it's going to be a significant number of changes.
- Add linters only, without breaking the build on failure. This should be a temporary solution. Follow up PRs should introduce the changes piece by piece. Then we can force linters.

munish7771: oh, didnt see the Work in progress label.

ifoukarakis:
> Are the changes in files test* because of linter?

Please see the other comment.

tests/utils/test_run_dbt_cloud_job.py
catalintomai: I'd say let's use the second approach. Linting is not urgent and/or breaking.

catalintomai (approved):

tests/utils/test_run_dbt_cloud_job.py
ifoukarakis: Ok, going for running but not failing approach. We can gradually move towards the first approach by requiring linters to pass on files that have changes.

PR #1062, opened 2022-11-21
MM-47904: Test for post feedback data

munish7771:
#### Summary

Contains unit tests for utils/post_docs_data.

- [x] refactored method names in conftest to be reused in test_post_docs_data.
- [x] refactored tests in test_post_nps_data.py with minor changes.
- [x] added unit tests for post_docs_data.py
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-47904

### DIFF
--- a/tests/utils/test_post_docs_data.py
+++ b/tests/utils/test_post_docs_data.py
@@ -1,0 +1,49 @@
+import os
+import importlib
+import pytest
+import snowflake.connector
+from utils import post_docs_data
+import responses
+
+class TestPostDocsJob():
+
+    @pytest.mark.parametrize("input_row, output_row",
+                                [
+                                    (('sample feedback 1',''),('sample feedback 1','')),
+                                    (('sample\nfeedback2',''),('sample feedback2','')),
+                                    (('sample\n\nfeedback ""3',''),('sample  feedback 3','')),
+                                    (('sample feedback""""\n""""4',''),('sample feedback 4','')),
+                                    (('sample feedback""""\n""""5','untouched string goes here\n\n""'),('sample feedback 5','untouched string goes here\n\n""'))
+                                ]
+                            )
+    def test_format_row(self, config_feedback, input_row, output_row):
+        # function returns output after removing newline and string quotes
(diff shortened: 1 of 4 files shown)


tests/utils/test_post_docs_data.py
```diff
@@ -0,0 +1,49 @@
...
+        assert post_docs_data.format_row(input_row) == output_row
+    
+    def test_feedback_generate_success(self, config_feedback, mock_snowflake_connector, mocker):
+
+        mock_snowflake_connector('utils.post_docs_data')
+        requests_mock = mocker.patch('utils.post_docs_data.requests.post')
+        requests_mock.return_value = mocker.Mock()
+        # Expected to give error since mock object methods cant be called
+        with pytest.raises(ValueError):
+            post_docs_data.post_docs()
```
ifoukarakis, line 30: Letting things fail because of incomplete mocking is not a good pattern. A better approach would be to just pass properties as keyword arguments.
```
        requests_mock.return_value = mocker.Mock(status_code=200)
```

However using `responses` seems like a better approach.

ifoukarakis (changes requested):

tests/utils/test_post_docs_data.py
```diff
@@ -0,0 +1,49 @@
+import os
+import importlib
+import pytest
+import snowflake.connector
+from utils import post_docs_data
+import responses
+
+class TestPostDocsJob():
```
catalintomai, line 8: @munish7771 , adding some comments regarding the purpose of these tests would be very useful. That way one does not have to jump to the original github PR to get the context (and it might not be always available)

munish7771:
@catalintomai @ifoukarakis 
Creating a new PR here due to merge conflicts  -> https://github.com/mattermost/mattermost-data-warehouse/pull/1121
Review changes have been addressed there.

tests/utils/test_post_docs_data.py
munish7771: Removed this test case. Implemented using responses.

PR #1088, opened 2022-12-01
MM-48455 Stitch alerting

munish7771:


dags/general/monitoring.py
```diff
@@ -0,0 +1,133 @@
...
+"""
+def stitch_check_extractions(response):
+    task_logger.info('Got extractions response from stitch api, checking for errors')
+    failed_extractions={}
+    try:
+        extractions = json.loads(response)['data']
+        for extraction in extractions:
+            if extraction['tap_exit_status'] is 1:
+                failed_extractions[extraction['source_id']] = extraction['tap_description']
+    except Exception as e:
```
ifoukarakis, line 29: It's not advised to catch `Exception` (or even have bare `except`). More details are available on this [post](https://faun.pub/python-exception-handling-anti-patterns-6e42aacff251)  and on [this article](https://realpython.com/the-most-diabolical-python-antipattern/), as well as suggestions on alternatives.

dags/general/monitoring.py
```diff
@@ -0,0 +1,133 @@
...
+        task_logger.error('error in getting extractions from stitch ', str(e))
```
ifoukarakis, line 30: A better approach would be to use `task_logger.exception` or `task_logger.error("error in ...", exc_info=True)`. This method automatically adds exception information to the log message. See [logging documentation](https://docs.python.org/3/library/logging.html#logging.Logger.exception).

dags/general/monitoring.py
```diff
@@ -0,0 +1,133 @@
...
+        raise(e)
```
ifoukarakis, line 31: Log and reraise is generally avoided. One of the main reasons is that it's easy for the exception to be logged elsewhere, wasting resources.

[This article](https://www.loggly.com/blog/exceptional-logging-of-exceptions-in-python/) describes some of the most common practices for handling exceptions.

dags/general/monitoring.py
```diff
@@ -0,0 +1,133 @@
...
+tap_exit_status = None, extraction in process
+"""
+def stitch_check_extractions(response):
+    task_logger.info('Got extractions response from stitch api, checking for errors')
+    failed_extractions={}
+    try:
+        extractions = json.loads(response)['data']
+        for extraction in extractions:
+            if extraction['tap_exit_status'] is 1:
+                failed_extractions[extraction['source_id']] = extraction['tap_description']
```
ifoukarakis, line 28: This code can be simplified a bit using dictionary manipulation.
```
# Safely get extractions, with an empty list as a default value
extractions = json.loads(response).get('data', []) 
failed_extractions = {i['source_id']: i['tap_description'] for i in extractions if i['tap_exit_status'] == 1}
```

dags/general/monitoring.py
```diff
@@ -0,0 +1,133 @@
...
+        raise(e)
+    return failed_extractions
+
+
+"""
+This method returns object of loads that are failing
+error_state != None, failing
+error_state = None, success
+"""
+def stitch_check_loads(response):
```
ifoukarakis, line 40: See comments in `stitch_check_extractions`.

dags/general/monitoring.py
```diff
@@ -0,0 +1,133 @@
...
+        task_logger.error('error in getting loads from stitch ', str(e))
+        raise(e)
+    return failed_loads
+
+"""
+This method fetches failed stitch extractions from xcom
+triggers a mattermost alert in case of failure
+""" 
+def resolve_stitch(**kwargs):
+    ti = kwargs['ti']
```
ifoukarakis, line 58: Since this function is expecting `ti` as argument, it's a good idea to define it. Function definition could be:

```
def resolve_stitch(ti=None, **kwargs):
```

dags/general/monitoring.py
```diff
@@ -0,0 +1,133 @@
...
+default_args={
+    'on_failure_callback': send_alert
+}
+
+dag = DAG('monitoring', 
+        default_args=default_args,
+        start_date=datetime(2017, 3, 20), 
+        catchup=False)
+
+with dag:
```
ifoukarakis, line 90: Not sure if `with dag` is going to work for the current version of airflow. If yes, one improvement is defining the dag as part of the `with` statement:

```
with DAG('monitoring',  default_args=default_args, start_date=datetime(2017, 3, 20),  catchup=False) as dag:
    ....
```

dags/general/monitoring.py
```diff
@@ -0,0 +1,133 @@
...
+                http_conn_id="stitch",
+                method="GET",
+                endpoint=Variable.get('stitch_extractions_endpoint'),
+                headers = {
+                        'Content-Type': 'application/json',
+                        'Authorization': Variable.get('stitch_secret')
+                        },
+                response_check=lambda x: True,
+                xcom_push=True,
+                dag=dag,
```
ifoukarakis, line 103: If `with` statement works for this version of airflow, then `dag` is injected automatically and shouldn't be needed here.

dags/general/monitoring.py
```diff
@@ -0,0 +1,133 @@
...
+                python_callable=resolve_stitch)
+
+    clean_xcom = PythonOperator(
+                task_id="clean_xcom",
+                python_callable = cleanup_xcom,
+                provide_context=True, 
+                dag=dag
+)
+
+check_stitch_extractions >> check_stitch_loads
```
ifoukarakis, line 131: Can `check_stitch_extractions` and `check_stitch_loads` be run in parallel?

ifoukarakis (commented):
Looks really promising! Added a few comments regarding python patterns/anti-patterns, plus some questions regarding the dag structure.

.github/workflows/ci-airflow.yml
```diff
@@ -39,6 +39,9 @@ jobs:
       - name: Install dependencies
         run: pip install -r build/requirements-airflow-dev.txt
 
+      - name: Initialize airflow database for loading variables
```
ifoukarakis, line 42: Adding this extra step will make the tests being non self-contained. There are a couple better approaches, please see `test_monitoring.py`.

dags/tests/general/test_monitoring.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+from airflow.models import Variable
+
+Variable.set("stitch_secret", "test_secret_value")
```
ifoukarakis, line 4: This requires that airflow backend have been configured and initialized. There are two approaches that can be used:
- Set the variables as env vars before running the tests (i.e. `AIRFLOW_VAR_STITCH_SECRET=test_secret_value AIRFLOW_VAR_STITCH_LOADS_ENDPOINT=test_loads_endpoint AIRFLOW_VAR_STITCH_EXTRACTIONS_ENDPOINT=test_extractions_endpoint pytest dags`). However this will need to be documented, as well as update `Makefile` and build.
- Mock `Variable`. Sample code would look like:
```
def mock_get(key, *args, **kwargs):
    return { "stitch_secret": "test_secret_value", ...}.get(key)
    
def test...(load_data, mocker):
    mocker.patch.object(Variable, 'get', new=mock_get)
```
- The previous solution would not work as is because `Variable.get` is used in `monitoring.py` and will be called at module's loading time. Reason is that DAG creation is defined on module level. One improvement is to use the methods under test to a separate module (i.e. `_helpers.py`).
- A simpler combination of the above solutions would be to move methods under test to `_helpers.py` and then mock environment variables by adding them in `config_utils` at `conftest.py`.

dags/general/monitoring.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+"""
+This method fetches failed stitch extractions from xcom
+triggers a mattermost alert in case of failure
+"""
+
+
+def resolve_stitch(ti=None, **kwargs):
+    ti = kwargs['ti']
```
ifoukarakis, line 83: No need for this line.

dags/general/monitoring.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+default_args = {'on_failure_callback': send_alert}
+
+with DAG(
+    'monitoring',
+    default_args=default_args,
+    start_date=datetime(2017, 3, 20),
+    schedule_interval='@hourly',
+    catchup=False,
```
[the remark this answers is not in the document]
munish7771: send alert is configured in default_args.

dags/general/monitoring.py
```diff
@@ -0,0 +1,147 @@
...
+        loads = json.loads(response)
+        if ('data' not in loads) or len(loads) == 0:
+            raise StitchApiException('Invalid response from loads api')
+        failed_loads = {
+            load['source_name']: load['error_state']['notification_data']['error']
+            for load in loads.get('data')
+            if load['error_state'] is not None
+        }
+    except KeyError:
+        task_logger.error('Error in check loads ...', exc_info=True)
```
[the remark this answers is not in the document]
munish7771: re throwing the exceptions in stitch_check_loads and stitch_check_extractions.

PR #1094, opened 2022-12-02
Hotfix: always pull latest image

ifoukarakis:
#### Summary

Image pull policy was set to `IfNotPresent`.  This resulted in the latest built image not being pulled. It has to be replaced with `Always` in order for latest changes to be visible.


munish7771 (approved):

spirosoik (approved):
LGTM. As we discussed after we make some improvements to airflow code, then it will be easier to support pinned versions instead of latest as a best practice.

PR #1097, opened 2022-12-07
Hotfix: deduplicate-server-security-details

ifoukarakis:
#### Summary

Incremental loading logic is using a `>=` statement for checking latest date. If the latest date happened to be the duplicate, then the new data would be created.

Since incremental loading requires newer data, it makes sense to include all records with timestamp greater (instead of greater or equal) than the max timestamp.


ifoukarakis:
Closing as increment logic is on date, not on timestamp.

PR #1100, opened 2022-12-09
MM-48977:  Introduce missing data check

ifoukarakis:
#### Summary

Introduces a new model for testing for missing activity data. This PR contains the following items:
- [x] DBT model for figuring out missing activity data per date. Last 90 days are used as "retention" period in order to save disk space.
- [x] DBT test checking the percentage of missing data for the past week against a configured threshold. 
- [x] Use a separate `data-quality` tag both for model and tests. This way data quality models and tests can be isolated and run independently of other models.

Note that thresholds can (and should) be changed. 

If you want to run the examples, you can try the following commands:
```bash
# Create model with historical missing data
dbt run --select tag:data-quality -t prod

# Run the test - requires the previous command
dbt test --select tag:data-quality -t prod
```

You can preview the output by checking `analytics.data_quality.dq_daily_activity` table.

Next step would be to run the tests daily, ideally after the daily DBT job has completed.

### DIFF
--- a/transform/snowflake-dbt/models/data_quality_layer/raw/rudderstack/dq_daily_activity.sql
+++ b/transform/snowflake-dbt/models/data_quality_layer/raw/rudderstack/dq_daily_activity.sql
@@ -1,0 +1,63 @@
+{{config({
+    "materialized": 'table',
+    "schema": "data_quality",
+    "tags":'data-quality'
+  })
+}}
+
+-- Number of days to keep activity data for monitoring
+{% set retention = 90 %}
+
+WITH server_activities AS (
+    -- Find the dates where a server has been sending activity data
+    SELECT
+        USER_ID AS server_id,
+        TIMESTAMP::DATE AS date,
+        COUNT(*) AS cnt_activities
+    FROM
+        {{ source('mm_telemetry_prod', 'activity') }}
+    WHERE
+        TIMESTAMP BETWEEN DATEADD('day', -{{ retention }}, CURRENT_TIMESTAMP) AND CURRENT_TIMESTAMP
+    GROUP BY 1, 2
+),
+server_activity_range AS (
+    -- Find the date range where each server has been sending data
+    SELECT
+        server_id,
+        MIN(date) AS start_date,
+        MAX(date) AS end_date
+    FROM
+        server_activities
+    GROUP BY 1
+),
+dates AS (
+    -- Find all dates between start and today
+    SELECT
+        DISTINCT(TIMESTAMP::DATE) AS date
+    FROM
+        {{ source('mm_telemetry_prod', 'activity') }}
+    WHERE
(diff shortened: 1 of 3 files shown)


catalintomai:
@ifoukarakis - wrt to "Next step would be to run the tests daily, ideally after the daily DBT job has completed" - can you set the schedule up with this PR or do you need a different one? I'd say the sooner we run this the better. 
I also wonder if we need to call it out in the path structure that this is for "activity" data and we might need similar ones for other types of events/actions (unless above we do the implicit mapping rudderstack->activity, which I'm not sure it is valid).

ifoukarakis:
> @ifoukarakis - wrt to "Next step would be to run the tests daily, ideally after the daily DBT job has completed" - can you set the schedule up with this PR or do you need a different one? I'd say the sooner we run this the better. I also wonder if we need to call it out in the path structure (e.g. ".../rudderstack/activity/.." that this is for "activity" data and we might need similar scripts for other types of data - e.g. "event", etc. Also, @sri-byte , do you think we need similar sets of scripts (or augment the ones in this PR) for mm_telemetry_qa/rc, other schemas?

Unfortunately I'll need a second PR. The reason is that this PR needs to be merged first in order to setup a job in Cloud DBT. Then the following PR will introduce the stack. 

I was also thinking about the granularity on path structure. My short answer is to keep it as is and if we see that we need some grouping based on input table, then we can revise. A longer answer is that I checked different strategies for DBT project structure. [DBT docs](https://docs.getdbt.com/guides/best-practices/how-we-structure/1-guide-overview) propose a flatter structure than our models, with the starting directory representing the layer, then going into the datasource etc. Unfortunately this post by DBT doesn't describe how scaling can work. I found [this medium post](https://medium.com/photobox-technology-product-and-design/practical-tips-to-get-the-best-out-of-data-building-tool-dbt-part-1-8cfa21ef97c5) goes in great detail about different approaches. All these approaches look interesting, but in the end it's a team decision.

catalintomai:
@ifoukarakis , makes sense, thanks. Wrt to the path structure, yes, let's revisit at a later time, just wanted us to think globally/holistically when we add point solutions.

transform/snowflake-dbt/models/data_quality_layer/raw/rudderstack/dq_daily_activity.sql
```diff
@@ -0,0 +1,63 @@
+{{config({
+    "materialized": 'table',
+    "schema": "data_quality",
+    "tags":'data-quality'
```
munish7771, line 4: Are we creating a new dbt cloud job that runs tag data-quality?
or we can also add it to dbt-nightly if its meant to run once per day.

munish7771 (approved):
+1 on test, Looks good !

PR #1109, opened 2022-12-14
MM-49037: reenable automatic deployments for python code

ifoukarakis:
#### Summary

- [x] Created a new DAG with `image_pull_policy` to `Always`. This DAG will pull the latest image with `master` tag (if any) every 5 minutes. 
- [x] Make sure tasks do not start if previous have not completed.
  - [x] Set DAG `max_active_runs` to `1` for all DAGs that use the `mattermost-data-warehouse` image.
  - [x] Moved image definition to `airflow_utils.py` so that it's defined on a single place.
  - [x] Properly moved `catchup` configuration as an argument to DAG's constructor rather than in `default_args`. [Airflow API](https://airflow.apache.org/docs/apache-airflow/1.10.12/_api/airflow/models/index.html?highlight=baseoperator#airflow.models.DAG) defines it this way.
- [x] Minor removals of dead code.

### DIFF
--- a/dags/airflow_utils.py
+++ b/dags/airflow_utils.py
@@ -102,7 +102,7 @@ pod_resources = Resources(request_memory="500Mi", request_cpu="500m")
 # Default settings for all DAGs
 pod_defaults = dict(
     get_logs=True,
-    image_pull_policy="Always",
+    image_pull_policy="IfNotPresent",
     in_cluster=True,
     is_delete_operator_pod=True,
     namespace=os.environ["NAMESPACE"],
--- a/dags/airflow_utils.py
+++ b/dags/airflow_utils.py
@@ -8,13 +8,12 @@ from typing import List
 from airflow.contrib.kubernetes.pod import Resources
 
 from plugins.operators.mattermost_operator import MattermostOperator
-from airflow.models import Variable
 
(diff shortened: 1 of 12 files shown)


dags/airflow_utils.py
```diff
@@ -103,7 +102,7 @@ def partitions(from_date: date, to_date: date, partition: str) -> List[dict]:
 # Default settings for all DAGs
 pod_defaults = dict(
     get_logs=True,
-    image_pull_policy="IfNotPresent",
+    image_pull_policy="Always",
```
munish7771, line 105: can we do something like this? only pull when a new version is available.
https://gist.github.com/byrnedo/9b2078c191360c681f85cebb2187d66f/revisions

ifoukarakis: Adding a bit more on my thinking, I'd expect that as soon as we finish with the current cleaning, changes on the python code would become slower. Then we can introduce some kind of versioning and use semver tags. As soon as this is in place, we can move from the current solution to the second approach.

munish7771: understood, we can do something like a mix of hack and SHA hash. 
Looks good for now.

munish7771 (approved):

munish7771 (approved):

PR #1116, opened 2022-12-16
Chore: update tests/utils code according to lint rules

ifoukarakis:
#### Summary

Fix linter errors under `tests/utils`. No change in the logic, mostly indentation and import order changes. Goal is to use single lint style for all code in order to improve readability.


munish7771 (approved):
lgtm!

PR #1123, opened 2022-12-20
MM-49070: introduce missing data checks for clearbit data

ifoukarakis:
#### Summary

Introduce missing data checks for cloud & on prem clearbit.

- [x] Add missing sources for clearbit data.
- [x] Add models to aggregate stats for servers enriched with clearbit data.
- [x] Add tests.

### DIFF
--- a/transform/snowflake-dbt/models/data_quality_layer/analytics/clearbit/dq_cloud_clearbit.sql
+++ b/transform/snowflake-dbt/models/data_quality_layer/analytics/clearbit/dq_cloud_clearbit.sql
@@ -1,0 +1,46 @@
+{{config({
+    "materialized": "table",
+    "schema": "data_quality",
+    "tags": "data-quality"
+  })
+}}
+
+
+WITH sso_installations AS (
+    SELECT
+        installation_id
+    FROM
+        {{ref('cs_signup_campaign_facts')}}
+    WHERE
+        sso_provider IS NOT NULL
+),
+sso_cloud_servers AS (
+    SELECT
+      sf.server_id
+      , sf.installation_id
+      , SPLIT_PART(license_email, '@', 2) AS email_domain
+      , COALESCE(sf.first_active_date, CURRENT_DATE) AS first_active_date
+      , sf.last_ip_address
+    FROM
+        {{ref('server_fact')}} sf
+        JOIN {{ref('license_server_fact')}} lsf ON sf.server_id = lsf.server_id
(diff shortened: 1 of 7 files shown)


transform/snowflake-dbt/models/data_quality_layer/analytics/clearbit/dq_cloud_clearbit.sql
```diff
@@ -0,0 +1,46 @@
...
+    i.installation_id,
+    cs.first_active_date,
+    cs.email_domain,
+    CASE WHEN cb.server_id IS NULL THEN 0 ELSE 1 END AS is_enriched,
+    -- Expect either company or person id to be there. If both are missing, then no data should be there.
+    CASE WHEN COALESCE(cb.company_id, cb.person_id) IS NULL THEN 1 ELSE 0 END AS is_missing
+FROM
+    sso_installations i
+    LEFT JOIN sso_cloud_servers cs ON i.INSTALLATION_ID = cs.INSTALLATION_ID
+    LEFT JOIN {{source('mattermost', 'cloud_clearbit')}} CB ON cs.server_id = cb.server_id
```
catalintomai, line 46: nit: CB -> cb

transform/snowflake-dbt/models/data_quality_layer/analytics/clearbit/dq_on_prem_clearbit.sql
```diff
@@ -0,0 +1,51 @@
+{{config({
+    "materialized": "table",
+    "schema": "data_quality",
+    "tags": "data-quality"
+  })
+}}
+
+
+WITH on_prem_servers AS (
```
catalintomai, line 9: @ifoukarakis - is there a way to factor out these ("sso_cloud_servers", "on_prem_servers", etc.) into some kind of utils set/library, etc. that we can include so that we can reuse them in multiple scripts. I can see them being used in many other places in the future.

catalintomai (approved):

transform/snowflake-dbt/models/data_quality_layer/analytics/clearbit/dq_on_prem_clearbit.sql
ifoukarakis: IMHO a better approach would be to have the "type" of the server in the `server_fact` table. This would make things simpler.

Another approach would be to move the related parts to a separate model. However I searched for similar occurrences in the project and didn't find something. Let's keep it in mind for now. Perhaps we can create a ticket to revise the `server_fact` table creation.

PR #1130, opened 2023-01-03
MM-49044: hightouch alerting

munish7771:
#### Summary

Added task in dag "monitoring" that uses HttpOperator to call hightouch API. Posts error in mattermost channel in case of failed syncs.

- [x] Adding methods for hightouch alerting in _helpers.py
- [x] Adding tasks to dag monitoring.py
- [x] Unit tests

Deployment steps -
- [x] Add airflow environment variables for -
1. hightouch_syncs_endpoint
2. hightouch_secret

- [x] Generate hightouch API key for a bot account. To be added to "hightouch_secret" above.
Looks like hightouch API keys are created at account level i.e mattermost, so a bot user is not needed.

TODO - 
1. Filter on hightouch syncs that need to be monitored. (https://mattermost.atlassian.net/browse/MM-48975)
2. Caching for hightouch alerts so that same alerts are not generated every hour.(https://mattermost.atlassian.net/browse/MM-48974)

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49044


sri-byte (approved):

dags/general/monitoring.py
```diff
@@ -57,6 +51,17 @@ def cleanup_xcom(**context):
...
+        task_id="check_hightouch_syncs",
+        http_conn_id="hightouch",
+        method="GET",
+        endpoint=Variable.get('hightouch_syncs_endpoint'),
+        headers={'Content-Type': 'application/json', 'Authorization': Variable.get('hightouch_secret')},
+        xcom_push=True,
+    )
+    resolve_hightouch_status = PythonOperator(
+        task_id='resolve_hightouch_status', provide_context=True, python_callable=resolve_hightouch
+    )
```
ifoukarakis, line 64: The two operators for hightouch are not wired together.

dags/general/_helpers.py
```diff
@@ -83,3 +88,46 @@ def resolve_stitch(ti=None, **kwargs):
...
+    except KeyError as e:
+        task_logger.error('Error in check syncs ...', exc_info=True)
+        raise e
+    except HightouchApiException as e:
+        task_logger.error('Error in ...', exc_info=True)
+        raise e
+    return failed_syncs
+
+
+def resolve_hightouch(ti=None, **kwargs):
```
ifoukarakis, line 116: This method is not tested.

dags/general/_helpers.py
```diff
@@ -83,3 +88,46 @@ def resolve_stitch(ti=None, **kwargs):
...
+    failed_syncs = hightouch_check_syncs(syncs)
+
+    if len(failed_syncs) == 0:
+        task_logger.info('There are no failed syncs')
+    else:
+        status = ':red_circle:'
+        message = f"**HIGHTOUCH**: {status}\nFailed syncs: {failed_syncs}"
+        MattermostOperator(mattermost_conn_id='mattermost', text=message, task_id='resolve_hightouch_message').execute(
+            None
+        )
```
ifoukarakis, line 133: This code is repeating in stitch monitoring as well. Perhaps we can have an operator that gathers all failures and aggregates them or move to a separate function.

ifoukarakis (changes requested):
There are missing tests and the operators are not wired together.

dags/general/monitoring.py
munish7771: setting downstream

ifoukarakis (approved):

PR #1132, opened 2023-01-09
MM-49444: document user event telemetry lineage

ifoukarakis:
#### Summary

- [x] Add missing sources in DBT `sources.yml`
- [x] Add links to the code that generates telemetry data.
- [x] Tag sources with the name of the repo(s) that send the data.


sri-byte:
why do we need these sources? Looks good otherwise.

sri-byte (approved):

ifoukarakis:
> why do we need these sources? Looks good otherwise.

These sources are used by models downstream but not documented. I added a bit more information to indicate the source of the data.

PR #1135, opened 2023-01-12
Changing user_events_telemetry job to run in nightly schedule

sri-byte:
Impact: This PR changes user_events_telemetry job to run in nightly schedule. I have identified that loading the table and selecting from this large table isn't running any queries at all. We aren't able to do any type of analysis and its holding back many Analytics tickets, that I am currently working on. We would not require this near real time data for Analytics, if there is ever a requirement to view real time data we could use the individual tables, for example - one scenario I can imagine where we would need real time data is when we want to see our own information on the Portal, say I want to see the flow of a user when signing up to a Cloud workspace. For this, we could just use the portal events table and it would be much more faster even.

Testing: Yesterday the job failed due to a column missing, and the job kept failing until I resolved the column issue. That was the only time I was able to run my query and progress on my work.

### DIFF
--- a/transform/snowflake-dbt/models/events/user_events_telemetry.sql
+++ b/transform/snowflake-dbt/models/events/user_events_telemetry.sql
@@ -1,7 +1,7 @@
 {{config({
     "materialized": "incremental",
     "schema": "events",
-    "tags":"union",
+    "tags":"nightly",
     "snowflake_warehouse": "transform_l"
   })
 }}


ifoukarakis (approved):

sri-byte:
Thank you @ifoukarakis. Lets me try to run the queries now.

2 main points to consider here, to not revert this -
-The queries I'm running have to execute
-Minimal to no impact on Nightly job.

transform/snowflake-dbt/models/events/user_events_telemetry.sql
```diff
@@ -1,7 +1,7 @@
 {{config({
     "materialized": "incremental",
```
catalintomai, line 2: @sri-byte , does it make sense to keep having this config setting here since we don't seem to be using it?

PR #1136, opened 2023-01-12
MM-49553: first version of precomputed user events telemetry

ifoukarakis:
#### Summary

Precomputed version of user events telemetry. Most calculations done in looker have been moved to this model. The following changes have been performed:
- All dimensions based on fields **ONLY** from `user_events_telemetry` have been moved to this table.
- Kept only fields appearing in Looker dimensions.
- Added `event_date` field to capture date of timestamp. This is also used as the clustering key since it's really often used in `WHERE` statement.
- Added incremental update. It's a bit tricky and not very performant to capture at this level, but it should not be a blocker ATM.

There's a lot of things that can be improved (converting to a proper fact table with dimensions, surrogate key etc), but there are improvements that can be done as soon as we validate the performance improvements.

### DIFF
--- a/transform/snowflake-dbt/models/events/user_events_telemetry_precalc.sql
+++ b/transform/snowflake-dbt/models/events/user_events_telemetry_precalc.sql
@@ -1,0 +1,241 @@
+{{config({
+    "materialized": "incremental",
+    "schema": "events",
+    "tags":"nightly",
+    "snowflake_warehouse": "transform_l",
+    "cluster_by": ['event_date']
+  })
+}}
+
+-- User events telemetry table, but with precalculated fields. This is an attempt to temporarily improve the
+-- performance of user_events_telemetry table. A better approach should consider the whole user event telemetry flows.
+
+WITH
+{% if is_incremental() %}
+max_time AS (
+    SELECT
+        _dbt_source_relation2
+        , MAX(received_at) as max_time
+    FROM
+        {{ this }}
+    GROUP BY 1
+),
+{% endif %}
+user_events_telemetry_precalc AS (
+    SELECT
+        uet.option AS option
+        , uet.campaign_utm_source AS campaign_utm_source
+        , uet.campaign_utm_medium AS campaign_utm_medium
+        , uet.campaign_utm_campaign AS campaign_utm_campaign
+        , uet.caller_info AS caller_info
+        , CASE
+            WHEN COALESCE(uet.type, uet.event) = 'pageview_getting_started' THEN COALESCE(uet.type, uet.event) || '_cloud'
+            ELSE COALESCE(uet.type, uet.event)
+          END AS type
+        , CASE WHEN type = 'api_reaction_save' THEN TRUE ELSE FALSE END AS reaction_event
+        , CASE WHEN type = 'LHS_DM_GM_Count' THEN TRUE ELSE FALSE END AS lhs_event
+        , uet.gaexp AS gaexp
+        , CASE WHEN uet.category = 'admin_team_config_page' THEN TRUE ELSE FALSE END AS admin_team_configuration_event
+        , CASE WHEN uet.category = 'admin_channel_config_page' THEN TRUE ELSE FALSE END AS admin_channel_configuration_event
+        -- The original query didn't define the `= 'cloud'` part. Without this part, the query is failing at snowflake
+        , CASE WHEN split_part(uet.category, '_', 1) = 'cloud' THEN TRUE ELSE FALSE END AS cloud_event
+        -- Following rows are results from joining other tables. If they are impacting performance, we can add them as part
+        -- of the precalculation process.
+        -- , CASE WHEN (split_part(${category}, '_', 1) = 'cloud') or (${server_fact.cloud_server} = TRUE) THEN TRUE ELSE FALSE END AS cloud_server
+        -- , COALESCE(${customers.email}, ${portal_customers.email}) AS stripe_customer_email
+        -- , COALESCE(${subscriptions.cws_dns}, ${portal_subscriptions.cws_dns}) AS stripe_customer_dns
+        -- Do we really need this?
+        -- , object_construct(user_events_telemetry.*) AS properties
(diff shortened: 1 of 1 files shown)


transform/snowflake-dbt/models/events/user_events_telemetry_precalc.sql
```diff
@@ -0,0 +1,241 @@
...
+            WHEN COALESCE(uet.type, uet.event) IN ('cloud_signup_b_page_visit','cloud_signup_page_visit') THEN 'Sign-Up Page'
+            WHEN COALESCE(uet.type, uet.event) = 'click_start_trial' THEN 'Click Start Trial'
+            WHEN COALESCE(uet.type, uet.event) = 'pageview_verify_email' THEN 'Verify Email'
+            WHEN COALESCE(uet.type, uet.event) = 'enter_valid_code' THEN 'Enter Valid Code'
+            WHEN COALESCE(uet.type, uet.event) = 'pageview_create_workspace' AND uet.name = 'pageview_company_name' THEN 'Enter Company Name'
+            WHEN COALESCE(uet.type, uet.event) = 'workspace_name_valid' THEN 'Workspace Name Valid'
+            WHEN COALESCE(uet.type, uet.event) = 'workspace_provisioning_started' THEN 'Workspace Creation Started'
+            WHEN COALESCE(uet.type, uet.event) = 'workspace_provisioning_ended' THEN 'Workspace Creation Ended'
+            ELSE NULL
+          END AS sign_up_events_sequence
```
sri-byte, line 120: We can remove this field, as I am doing this calculation in Looker.

transform/snowflake-dbt/models/events/user_events_telemetry_precalc.sql
```diff
@@ -0,0 +1,241 @@
...
+            WHEN COALESCE(uet.type, uet.event) IN ('cloud_signup_b_page_visit','cloud_signup_page_visit') THEN 1
+            WHEN COALESCE(uet.type, uet.event) = 'click_start_trial' THEN 2
+            WHEN COALESCE(uet.type, uet.event) = 'pageview_verify_email' THEN 3
+            WHEN COALESCE(uet.type, uet.event) = 'enter_valid_code' THEN 4
+            WHEN COALESCE(uet.type, uet.event) = 'pageview_create_workspace' AND uet.name = 'pageview_company_name' THEN 5
+            WHEN COALESCE(uet.type, uet.event) = 'workspace_name_valid' THEN 6
+            WHEN COALESCE(uet.type, uet.event) = 'workspace_provisioning_started' THEN 7
+            WHEN COALESCE(uet.type, uet.event) = 'workspace_provisioning_ended' THEN 8
+            ELSE NULL
+          END AS sign_up_sequence
```
sri-byte, line 109: We can remove this field, as I am doing this calculation in Looker.

transform/snowflake-dbt/models/events/user_events_telemetry_precalc.sql
ifoukarakis: That's the idea, not to have to do the calculation in Looker.

sri-byte: No, this calculation is done in the visualization, after analysis. There is an option to add calculated field based on the output of the query, which is not part of the select statement.

sri-byte: We can drop the field in Looker, to avoid confusion.

sri-byte: the type and event we need to keep, the sign_up_events_sequence we can remove

transform/snowflake-dbt/models/events/user_events_telemetry_precalc.sql
```diff
@@ -0,0 +1,220 @@
...
+    "cluster_by": ['event_date']
+  })
+}}
+
+-- User events telemetry table, but with precalculated fields. This is an attempt to temporarily improve the
+-- performance of user_events_telemetry table. A better approach should consider the whole user event telemetry flows.
+
+WITH
+{% if is_incremental() %}
+max_time AS (
```
catalintomai, line 15: Is this `max_time` or `mt`?

ifoukarakis: `max_time` the name of the CTE. `mt` is the alias in the `JOIN` statement.

catalintomai: So there is no need to have a `JOIN make_table mt ...` then?

ifoukarakis: The role of `max_time` CTE is to enable incremental loading in the current table. It provides the timestamp of the latest record from the source table. The value is stored in column `max_time` of the CTE. The trick is that the source table is the union of multiple tables, so the latest timestamp from each of the original source tables is not the same value. For this reason, the latest record's timestamp is stored per each source table (`_dbt_source_relation2`).

When the actual incremental join happens, the source table is filtered by joining the original table with `max_time` CTE. `mt` is the alias of the CTE in this join. 

One thing that can be confusing is that both the CTE and column are called `max_time`. Perhaps the CTE can be renamed to `latest_records` and the column to `latest_received_at`.

sri-byte: should we use `timestamp` here, instead of `received_date`, since clustering index is on timestamp.

ifoukarakis: @sri-byte [Snowflake documentation](https://docs.snowflake.com/en/user-guide/tables-clustering-keys.html#strategies-for-selecting-clustering-keys) suggest to use columns that are most actively used in selective filters. Based on the queries provided, it seems that in the `WHERE` part of the queries the timestamp is cast to date. This version of the table builds on that, adding date as a separate column. Queries should be updated to utilize the `event_date` column rather than performing a cast.

catalintomai (approved):

PR #1144, opened 2023-01-19
MM-49552: add link to sources for segment data

ifoukarakis:
#### Summary

Added documentation for sources that originate from segment.


munish7771 (approved):
looks good!

PR #1152, opened 2023-01-27
MM-49894: First version of event registry

ifoukarakis:
#### Summary

First version of event registry. Iterates over all Rudderstack [tracks](https://www.rudderstack.com/docs/destinations/warehouse-destinations/warehouse-schema/#track) table and joins them across sources. 

Please note the following design choices:
- Ignoring all segment originating events for now.
- Each `tracks` table has its own `base` model. This allows more finegrained control on table level.  A side benefit of this is that new `tracks` can be added without requiring complex incremental logic. This also helped greatly in optimizing execution time.
- Base tables are updated incrementally to reduce resource usage.
- Repeatable code has been extracted to a macro. This will make it easier. 
- There were some missing table definitions for `tracks` tables in sources that needed to be added.

### DIFF
--- a/transform/snowflake-dbt/models/event_registry/event_registry.yml
+++ b/transform/snowflake-dbt/models/event_registry/event_registry.yml
@@ -1,0 +1,76 @@
+version: 2
+
+models:
+  - name: base_tracks
+    description: >
+      Daily event information. Doesn't contain rows for events on dates where 
+      no events of that type were submitted.
+
+    columns:
+      - name: id
+        description: A unique id for each event/date/source combination.
+        tests:
+          - unique
+          - not_null
+      - name: date_received_at
+        description: The date that events of the current type were received.
+        tests:
+          - not_null
+      - name: event
+        description: The name of the corresponding event table.
+        tests:
+          - not_null
+      - name: event_text
+        description: The name of the event.
+        tests:
+          - not_null
+      - name: source
+        description: The name of the source that the event are submitted from.
+        tests:
+          - not_null
+      - name: event_count
+        description: The total number of events submitted for the current date.
+        tests:
+          - not_null
+
+  - name: event_registry
+    description: |
+      Provides a list of events captured using Rudderstack's 
+      [track](https://www.rudderstack.com/docs/event-spec/standard-events/track/) API call. 
+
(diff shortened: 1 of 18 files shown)


transform/snowflake-dbt/models/event_registry/event_registry.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: id
+        description: A unique id for each event/date/source combination.
+        tests:
+          - unique
+          - not_null
+      - name: date_received_at
+        description: The date that events of the current type were received.
+        tests:
+          - not_null
+      - name: event
```
catalintomai, line 19: @ifoukarakis , maybe just call it `event_table` and the one below `event` or `event_name`? `event_text` seems more appropriate for some event description field.

ifoukarakis: These are actually columns of the base tables. I reused the same descriptions for the columns as in https://www.rudderstack.com/docs/destinations/warehouse-destinations/warehouse-schema/#table-tracks. Am I missing something here?

We can also discuss names for `event_registry` and `daily_event_stats` tables. I thought of using `events`, but such table exists in other schemas, so I thought of avoiding conflicting names.

catalintomai: Yeah, not the best (initial) naming imo, but better be consistent in this case I guess.

ifoukarakis: @catalintomai what about the following:
- Rename `event_registry` to `event_fact`.
- Rename column `event` to `event_table` or `table`.
- Rename `event_text` to `event_name` to `name`.

Alternatively, we can use the naming above but with removing the `event_` prefix. 

`daily_event_stats` should keep the same name, but follow the same conventions.

catalintomai (approved):

transform/snowflake-dbt/models/event_registry/event_registry.yml
catalintomai: Thanks @ifoukarakis , that makes more sense/easier to follow to me (prefer `event_*` to `*` ).

PR #1154, opened 2023-01-31
MM-50113: Add tests for cloud professional subscriptions

munish7771:
#### Summary

Collects rows from the stage table and compares it with destination `ORGM`. Entity is confirmed as synced into salesforce if `sfid` exists for it.
- [x] Added basic tests for unique and not_null fields in schema.yml
- [x] Added tests for dql that run daily via airflow 

Deployment steps - N/A (assuming it runs via airflow nightly job because of tag `data-quality`)
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-50113


nat-gunner (approved):

ifoukarakis (approved):
🚀

PR #1159, opened 2023-02-10
MM-50437: setup new dbt project

ifoukarakis:
#### Summary

- [x] Setup new project structure according to [suggested practices](https://docs.getdbt.com/guides/best-practices/how-we-structure/1-guide-overview) by dbt.
- [x] Override schema name generation according to [alternative approach](https://docs.getdbt.com/docs/build/custom-schemas#an-alternative-pattern-for-generating-schema-names).
- [x] Include commonly used packages (dbt-utils and codegen). 
- [x] Update documentation.
- [x] Add `make` target for new version of DBT. Opted for a single target than separate targets for building docs, service etc (all can be done within the same container).
- [x] Modified prompts in docker environments so that
  - Time is visible as part of the prompt (this way we can quickly calculate how long an operation took).
  - The project (`snowflake-dbt` vs `mattermost-analytics`) is explicitly visible.
  - A nice reminder that we're running within docker using a 🐳  emoji.

Here's how it looks


sri-byte (approved):

sri-byte:
LGTM

PR #1170, opened 2023-02-15
MM-50377: add telemetry schemas

ifoukarakis:
#### Summary

Add `mm_telemetry_prod` and `mm_telemetry_rc` schemas in staging layer. Both schemas contain junk tables with a small number of rows. Rather than attempting to join all of them, a heuristic is used to determine which tables to merge. The rules are:
- Tables must be in the given schema.
- Tables must have a minimum number of rows (configurable).
- Table must not be one of the default rudderstack tables.

The information required for the decision is read from Snowflake's information schema. The drawback of this approach is that new events will need to reach the minimum number of rows before displayed. 

An alternative approach would be to get the distinct name of event tables from `tracks` table. However, this can be expensive, as it requires iterating the whole table.


transform/mattermost-analytics/dbt/sources.yml
```diff
@@ -0,0 +1,845 @@
+[0m12:25:13  Running with dbt=1.4.1
+[0m12:25:14  Unable to do partial parsing because config vars, config profile, or config target have changed
+[0m12:25:21  [[33mWARNING[0m]: Configuration paths exist in your dbt_project.yml file which do not apply to any resources.
+There are 3 unused configuration paths:
+- models.mattermost_analytics.marts
+- models.mattermost_analytics.utilities
+- models.mattermost_analytics.intermediate
+[0m12:25:26  version: 2
```
catalintomai, line 8: should we remove this paragraph?

ifoukarakis: Yes, you are right. I'll actually also add it in `.gitignore`.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__sources.yml
```diff
@@ -0,0 +1,200 @@
...
+      - name: _groups
+        description: Contains raw server-level group data as a single row per server per timestamp.
+
+      - name: channel_moderation
+        description: Contains raw server-level channel data as a single row per server per timestamp.
+
+      - name: elasticsearch
+        description: Contains raw server-level ElasticSearch data as a single row per server per timestamp.
+
+      - name: warn_metrics
```
catalintomai, line 178: I would be removing this. While we still receive telemetry from older versions, the telemetry is not used anymore, feature/functionality has been cut.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__sources.yml
```diff
@@ -0,0 +1,200 @@
...
+      - name: config_theme
+        description: The raw Server theme configuration data sent to Mattermost by Telemetry-Enabled Servers.
+
+      - name: config_timezone
+        description: The raw Server timezone configuration data sent to Mattermost by Telemetry-Enabled Servers.
+
+      - name: config_webrtc
+        description: The raw Server WebRTC configuration data sent to Mattermost by Telemetry-Enabled Servers.
+
+      - name: config_audit
```
catalintomai, line 131: should `config_audit` `config_bleve` be arranged alphabetically in the list of config_* tables above?

transform/mattermost-analytics/macros/rudderstack/_rudderstack_schema.yml
```diff
@@ -22,3 +22,17 @@ macros:
         type: list
         description: The list of columns to keep.
 
+  - name: get_event_relations
+    description: |
+      Returns a list of relations from the given schema that might be Rudderstack events. Offers functionality for
```
catalintomai, line 27: nit: `relations` -> `tables` (schemas do not contain/denote relations)

catalintomai, line 27: maybe mention that there relations are tables (and not other types of schema elts like views)

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__sources.yml
ifoukarakis: Perhaps there are other events like this. It would be easy to figure out as soon as we have the report in looker. I'd suggest to create a list of deprecated events as soon as we finish the current ticket, and then think whether we want to show them as deprecated on Looker or just remove them.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__sources.yml
ifoukarakis: I sorted the whole list of tables alphabetically and cross-checked if there are missing tables.

catalintomai (approved):

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__sources.yml
```diff
@@ -0,0 +1,214 @@
+version: 2
+
+sources:
+  - name: mm_telemetry_prod
+    database: 'RAW'
+    schema: mm_telemetry_prod
+    loader: Rudderstack
+    description: |
+      Mattermost server and webapp telemetry data. Stored using Rudderstack. Rudderstack documentations offers an
```
catalintomai, line 9: @ifoukarakis , wonder if we need to make it clear here or somewhere else that we only capture "more recent" events (Rudderstack) and not the older ones (Segment). Drawing the line at the product version where Rudderstack has been added (v5.0?) might convey the same.

PR #1172, opened 2023-02-16
MM-50377: Add intermediate, fact and dimension tables

ifoukarakis:
#### Summary

Introduce event registry fact & dimension tables. This design includes

- [x] Fact table for daily event snapshot.
- [x] Dimension table for events.
- [x] Intermediate tables.
- [x] Utility macros for reducing repeatable code in intermediate layer.
- [x] Docs & tests for fact & dimension table.
- [x] Minor fix on naming convention for `event_type` .

I did not include a `date` dimension for now, as we don't have a solid use case (yet). Since it's extremely likely we'll need it for other use cases as well, we can introduce it as an extra step.

### DIFF
--- a/transform/mattermost-analytics/models/marts/data_eng/event_registry/dim_events.sql
+++ b/transform/mattermost-analytics/models/marts/data_eng/event_registry/dim_events.sql
@@ -1,7 +1,6 @@
 {{
     config({
         "tags":"hourly",
-        "cluster_by": ['event_name', 'source'],
     })
 }}
 
--- a/transform/mattermost-analytics/models/marts/data_eng/event_registry/dim_events.sql
+++ b/transform/mattermost-analytics/models/marts/data_eng/event_registry/dim_events.sql
@@ -1,0 +1,16 @@
+{{
+    config({
+        "tags":"hourly",
+        "cluster_by": ['event_name', 'source'],
+    })
+}}
+
+SELECT DISTINCT
+    event_id
+    , event_name
+    , event_table
+    , category
+    , event_type
+    , source
+FROM
+    {{ ref('int_events_aggregated_to_date') }}
--- a/transform/mattermost-analytics/models/marts/data_eng/event_registry/fct_events_daily_snapshot.sql
+++ b/transform/mattermost-analytics/models/marts/data_eng/event_registry/fct_events_daily_snapshot.sql
@@ -6,7 +6,7 @@
 }}
 
 SELECT DISTINCT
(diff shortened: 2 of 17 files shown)


transform/mattermost-analytics/models/marts/data_eng/event_registry/fct_events_daily_snapshot.sql
```diff
@@ -0,0 +1,14 @@
+{{
+    config({
+        "tags":"hourly",
```
catalintomai, line 3: why is it a `daily` snapshot (given that the tags value is hourly)?

ifoukarakis: This tag is used to configure how often each model runs. Although it's a daily snapshot, it makes sense to update hourly so that we can see new events on the date they were added (with 1 hour delay top). 

If this is not needed, the tag can change to `nightly`.

transform/mattermost-analytics/models/marts/data_eng/event_registry/dim_events.sql
```diff
@@ -0,0 +1,16 @@
+{{
+    config({
+        "tags":"hourly",
+        "cluster_by": ['event_name', 'source'],
```
catalintomai, line 4: are there any pros/cons to cluster by `event_name` vs. `event_id`?

ifoukarakis: Event id is unique and a string. I'm not sure that it offers good granularity, as it doesn't really allow to group events in the dimension. 

TBH, given the current number of rows in this table, I'm not sure it will make an important difference. We can drop clustering for now.

catalintomai: Good point, wondering what clustering really means for dimension tables, or is it just a "smth. one can do" kind of thing.

ifoukarakis: I will remove it for now and we can reconsider if we see performance issues.

catalintomai (approved):

transform/mattermost-analytics/models/marts/data_eng/event_registry/dim_events.sql
ifoukarakis: Removed the clustering and fixed a bug in grouping.

PR #1173, opened 2023-02-17
Adding staging tables for performance_events

sri-byte:
Impact: Optimizing performance events, 

- Adding staging tables with columns only needed for Looker and removing other columns.


transform/mattermost-analytics/models/staging/mm_telemetry_rc/stg_mm_telemetry_rc__performance_events.sql
```diff
@@ -0,0 +1,73 @@
...
+ , method                         AS method
+ , remaining                      AS remaining
+ , screen                         AS screen
+ , filter                         AS filter
+ , version                        AS version
+ , original_timestamp             AS original_timestamp
+ , sent_at                        AS sent_at
+ , uuid_ts                        AS uuid_ts
+ , received_at                    AS received_at 
+ FROM {{ ref ('events') }} WHERE category = 'performance')
```
ifoukarakis, line 72:
```suggestion
 FROM {{ source ('mm_telemetry_rc', 'events') }} WHERE category = 'performance')
```
Models is staging layer must use either `source` models or base models in the staging layer.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
```diff
@@ -0,0 +1,73 @@
...
+ , method                         AS method
+ , remaining                      AS remaining
+ , screen                         AS screen
+ , filter                         AS filter
+ , version                        AS version
+ , original_timestamp             AS original_timestamp
+ , sent_at                        AS sent_at
+ , uuid_ts                        AS uuid_ts
+ , received_at                    AS received_at 
+ FROM {{ ref ('events') }} WHERE category = 'performance')
```
ifoukarakis, line 72:
```suggestion
 FROM {{ source ('mm_telemetry_prod', 'events') }} WHERE category = 'performance')
```
Models is staging layer must use either `source` models or base models in the staging layer.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
```diff
@@ -0,0 +1,73 @@
+{{
+    config({
+        "tags":"hourly",
+    })
+}}
+
+WITH MM_TELEMETRY_PROD AS 
+(SELECT _dbt_source_relation	  AS _dbt_source_relation
```
ifoukarakis, line 8: `_dbt_source_relation` is a column that is added by `dbt_utils.union_relations`. Table `events` doesn't include that.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
```diff
@@ -0,0 +1,73 @@
...
+ , id                             AS id
+ , user_actual_id                 AS user_actual_id
+ , context_library_name           AS context_library_name
+ , type                           AS type
+ , context_app_version            AS context_app_version
+ , user_actual_role               AS user_actual_role
+ , context_page_url               AS context_page_url
+ , context_os_name                AS context_os_name
+ , context_page_title             AS context_page_title
+ , anonymous_id                   AS anonymous_id
```
ifoukarakis, line 21: IIRC correctly, `anonymous_id` is always added with a default value. Perhaps it's not needed.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
```diff
@@ -0,0 +1,73 @@
+{{
+    config({
+        "tags":"hourly",
+    })
+}}
+
+WITH MM_TELEMETRY_PROD AS 
+(SELECT _dbt_source_relation	  AS _dbt_source_relation	
+ , context_page_referrer          AS context_page_referrer
+ , channel                        AS channel
```
ifoukarakis, line 10: Quite a few of these columns are shared across all Rudderstack events. It's extremely likely that we'll need them again when extracting data for other events. Perhaps we can find a way to abstract this so that we don't repeat it over code.

Adding a bit more info, each event's columns are divided in three groups:
- Standard rudderstack columns (i.e. `id`, `anonymous_id`, `timestamp`, `received_at` etc). A detailed list is available [here](https://www.rudderstack.com/docs/destinations/warehouse-destinations/warehouse-schema/#table-add_to_cart).
- Standard column conventions from Mattermost's servers. Examples are `user_id`, `user_actual_id`, `category`, `type`. The tricky part with these columns is they don't exist in every event table.
- Event-specific properties. 

The first two groups can be abstracted to a macro (i.e. `rudderstack_columns()`). This will increase readability of staging models.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
```diff
@@ -0,0 +1,73 @@
...
+ , context_locale                 AS context_locale
+ , context_screen_density         AS context_screen_density
+ , context_page_path              AS context_page_path
+ , context_os_version             AS context_os_version
+ , category                       AS category
+ , duration                       AS duration
+ , num_of_request                 AS num_of_request
+ , max_api_resource_size          AS max_api_resource_size
+ , longest_api_resource_duration  AS longest_api_resource_duration
+ , user_id                        AS user_id
```
catalintomai, line 37: Is there a place in the new codebase to document some of these columns? Like the difference between `user_id` and `userid`? cc: @ifoukarakis , @sri-byte

transform/mattermost-analytics/models/staging/mm_telemetry_rc/_mm_telemetry_rc__models.yml
```diff
@@ -21,4 +21,75 @@ models:
...
         description: The ID of the server the event originated from.
       - name: received_at
-        description: Timestamp registered by RudderStack when the event was ingested (received).
\ No newline at end of file
+        description: Timestamp registered by RudderStack when the event was ingested (received).
+
+        
+  - name: stg_mm_telemetry_rc__performance_events
+    description: |
+      Reconstructed `performance_events` table using only columns needed by looker.
```
catalintomai, line 29: Maybe document/mention the difference between `rc` and prod`, otherwise the descriptions are identical.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
catalintomai: more about `anonymousId`: https://www.rudderstack.com/docs/event-spec/standard-events/identify/

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
sri-byte: They're both the same, I suspect it was a bug during development that was later resolved. 0/5 on a place to document this.

ifoukarakis: We can coalesce them if that's the case. After all, the staging layer is the perfect place for performing cleanup on source data.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__models.yml
```diff
@@ -21,4 +21,74 @@ models:
       - name: server_id
         description: The ID of the server the event originated from.
       - name: received_at
-        description: Timestamp registered by RudderStack when the event was ingested (received).
\ No newline at end of file
+        description: Timestamp registered by RudderStack when the event was ingested (received).
+
+    - name: stg_mm_telemetry_prod__performance_events
```
catalintomai, line 26: why not add some generic tests at least for the "main" columns? (here and for `rc`)

transform/mattermost-analytics/macros/rudderstack/union_relations.sql
```diff
@@ -0,0 +1,128 @@
+{%- macro union_relations(relations, column_override=none, include=[], exclude=[], source_column_name='_dbt_source_relation', where=none) -%}
```
ifoukarakis, line 1: No need to add this. This macro originates from [dbt-utils](https://github.com/dbt-labs/dbt-utils) package. This package is already part of the project, so you can use the [union_relations](https://github.com/dbt-labs/dbt-utils#union_relations-source) macro from the package directly.

Here's [an example](https://github.com/mattermost/mattermost-data-warehouse/blob/master/transform/mattermost-analytics/models/intermediate/data_eng/event_registry/int_events_aggregated_to_date.sql) on how it's used in another filemodelin the project.

transform/mattermost-analytics/macros/rudderstack/get_rudderstack_columns.sql
```diff
@@ -0,0 +1,12 @@
+{% macro get_rudderstack_columns() %}
+
+   {%-
+        set rudderstack_columns = ['id', 'anonymous_id', 'received_at', 'sent_at', 'original_timestamp', 'timestamp', 
+        'context_ip', 'event', 'event_text']
+    -%}
+
+    {% for column in rudderstack_columns %}
+    , {{ column }} AS {{ column }}
+    {% endfor %}
```
ifoukarakis, line 10: Following the example [here](https://docs.getdbt.com/docs/get-started/learning-more/using-jinja#use-looplast-to-avoid-trailing-commas), the code can be improved to avoid leading commas.
```suggestion
    {% for column in rudderstack_columns %}
        {{ column }} AS {{ column }}
        {% if not loop.last %},{% endif %}
    {% endfor %}
```

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
```diff
@@ -0,0 +1,26 @@
...
+, "num_of_request", "max_api_resource_size", "longest_api_resource_duration", "user_id", "userid" 
+, "root_id", "post_id", "sort", "team_id", "version", "keyword", "count"
+, "gfyid", "field", "plugin_id" , "installed_version", "group_constrained", "value", "include_deleted" 
+, "role", "privacy", "scheme_id" , "warnmetricid"  , "metric" , "error"
+, "num_invitations_sent", "num_invitations" , "channel_sidebar" , "app" , "method" , "remaining" 
+, "screen" , "filter" , "version" , "uuid_ts", "id", "anonymous_id", "received_at", "sent_at", "original_timestamp", "timestamp", 
+        "context_ip", "event", "event_text"], where = " CATEGORY = 'performance'",
+  )}}
+
+SELECT * FROM MM_TELEMETRY_PROD
```
ifoukarakis, line 26: Using `dbt_utils.union_relations` here for a single table will do the work. However the name can be a bit confusing for anyone who reads the code. One thing that can improve the readability of the code is to change this to something like this (untested):
```suggestion
{%- set include_columns = ["context_page_referrer", "channel", "context_app_namespace", "user_actual_id", 
    "context_library_name", "type", "context_app_version" , "user_actual_role", "context_page_url"
, "context_os_name" , "context_page_title"  ,  "context_app_build", "context_page_search" 
, "context_library_version",  "context_useragent", "context_app_name", "context_locale"
, "context_screen_density" , "context_page_path", "context_os_version"  , "category"  , "duration"  
, "num_of_request", "max_api_resource_size", "longest_api_resource_duration", "user_id", "userid" 
, "root_id", "post_id", "sort", "team_id", "version", "keyword", "count"
, "gfyid", "field", "plugin_id" , "installed_version", "group_constrained", "value", "include_deleted" 
, "role", "privacy", "scheme_id" , "warnmetricid"  , "metric" , "error"
, "num_invitations_sent", "num_invitations" , "channel_sidebar" , "app" , "method" , "remaining" 
, "screen" , "filter" , "version" , "uuid_ts", "id", "anonymous_id", "received_at", "sent_at", "original_timestamp", "timestamp", "context_ip", "event", "event_text"] %-}
        
WITH performance_event AS (
    SELECT
      {{ get_rudderstack_columns() }},
        {% for column in include_columns %}
        {{ column }} AS {{ column }}
        {% if not loop.last %},{% endif %}
    {% endfor %}
    FROM
      {{{ source('mm_telemetry_prod', 'event') }}
    WHERE CATEGORY = 'performance'",
)

SELECT * FROM performance_events
```

This  code might seem a bit more, but it communicates clearly the intent of filtering some of the columns  from `event` table in order to keep only records where `category = 'performance'`.

There are also a few more improvements that can be done here. 
- `include_columns` contains a couple of columns that are already defined in `get_rudderstack_columns`, so they need to be removed from `include_columns`.
- `include_columns` contains a few columns that are reused across events in `events` table. For example `category`, `user_id`, `user_actual_id`, `user_actual_role`, `category`, `type` should be all present. These columns can be extracted to a separate list or a macro similar to `get_rudderstack_columns`. This way they will be reusable whenever a new staging model that extracts specific events from `events` is added.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
```diff
@@ -0,0 +1,31 @@
+{{
+config({
+ "tags":"hourly",
+})
+}}
+
+{%- set include_columns = ["context_page_referrer", "channel", "context_app_namespace", "user_actual_id"
```
ifoukarakis, line 7: Do we need all these columns? For example `uuid_ts` is used for debugging purposes (according to [Rudderstack's doc](https://www.rudderstack.com/docs/destinations/warehouse-destinations/warehouse-schema/)).

`context_*` fields is another category. For example `context_library_name` is the name of the rudderstack library.

transform/mattermost-analytics/models/intermediate/web_app/int_mm_telemetry_rc__performance_events.sql
```diff
@@ -0,0 +1,10 @@
+{{
+    config({
+        "materialized": "incremental",
+        "incremental_strategy": "merge",
+        "cluster_by": ['timestamp'],
+    })
+}}
+
+SELECT * FROM
```
ifoukarakis, line 9: Incremental logic needs to be added here as well.

transform/mattermost-analytics/models/intermediate/web_app/int__performance_events.sql
```diff
@@ -0,0 +1,9 @@
+{{
+    dbt_utils.union_relations(
+        relations=[
+            ref('int_mm_telemetry_prod__performance_events'),
+            ref('int_mm_telemetry_rc__performance_events'),
+        ],
+        source_column_name=None
```
ifoukarakis, line 7: Source would be nice to keep. The other examples I used already had a source field, so it would be duplicate information. That's the reason `source_column_name` is none for the event registry models.

transform/mattermost-analytics/models/intermediate/web_app/int_mm_telemetry_prod__performance_events.sql
```diff
@@ -0,0 +1,10 @@
+{{
+    config({
+        "materialized": "incremental",
+        "incremental_strategy": "merge",
+        "cluster_by": ['timestamp'],
```
ifoukarakis, line 5: [Snowflake documentation](https://docs.snowflake.com/en/user-guide/tables-clustering-keys#strategies-for-selecting-clustering-keys) has some suggestions on how to choose a clustering key. It is mentioned that `In general, if a column (or expression) has higher cardinality, then maintaining clustering on that column is more expensive.`. That's the reason clustering is usually done on a date rather than on a timestamp.

sri-byte: Added a new field in staging, which extracts date from `timestamp` and using that as cluster_key, which I plan to also use in Looker.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
sri-byte: I copied the columns that are in Looker, I'll check with @hmhealey tomorrow to see what fields they use from performance events, and we can probably drop most of the fields.

transform/mattermost-analytics/models/marts/web_app/platform/fct_events_daily_snapshot.sql
```diff
@@ -0,0 +1,8 @@
+{{
```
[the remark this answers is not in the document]
ifoukarakis: (nitpicking) since this table is not a real fact table, perhaps we can think a better name for the model file. 
I was thinking that since it's grouping events for `performance` category, we can use a prefix of `eg` (for event group). The rest can be something similar to `eg_[group name]s_[group table type]s.sql`. Group table type can be:
- Individual events (similar to individual transaction in a fact table).
- Periodic Snapshot (i.e. hourly, daily, weekly etc snapshot - similar to periodic snapshot fact)
- Accumulating snapshot (in case we want to track events order per user actions).

If I understand correctly, the current case is about individual events, so perhaps a nice model name would be `eg_performance_events`, assuming that the group name is `performance` and granularity is `event`.

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_prod_performance_events.sql
```diff
@@ -0,0 +1,13 @@
...
+        "materialized": "incremental",
+        "cluster_by": ['event_date'],
+    })
+}}
+
+SELECT * FROM 
+{{ ref('stg_mm_telemetry_prod__performance_events') }}
+
+{% if is_incremental() %}
+    WHERE event_date >= (SELECT MAX(event_date) FROM {{ this }})
```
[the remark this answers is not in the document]
sri-byte: The reason I used this, is because I want the table to be preformative, mainly in Looker. The cluster_key is crucial for this reason, I don't have a problem to use `received_at`, provided the cluster is on `event_date`. Across the board for most analyses we use `timestamp` or `original_timestamp`

transform/mattermost-analytics/models/marts/web_app/platform/fct_events_daily_snapshot.sql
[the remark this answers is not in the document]
sri-byte: > This model and `int_performance_events` can be merged. It can also be updated incrementally, but this might require replacing `union_relations` with manual `UNION ALL` statement.

What is the idea behind merging these, `fct_events_daily_snapshot` has only 4 columns. `performance_events` is already a pretty huge table in terms of columns and volume.

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_prod_performance_events.sql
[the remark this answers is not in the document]
catalintomai: Are the intermediate model tables used in Looker directly or only the fact/dim tables in the `marts` folder?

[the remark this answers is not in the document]
sri-byte: only tables from `marts` folder

transform/mattermost-analytics/models/staging/mm_telemetry_rc/stg_mm_telemetry_rc__performance_events.sql
```diff
@@ -0,0 +1,35 @@
+{{
+config({
+ "tags":"hourly",
```
catalintomai, line 3: do we still need this if we've set it globally in the `dbt_project.yml` file?

ifoukarakis: No, it can be removed.

transform/mattermost-analytics/models/marts/web_app/platform/fct_events_daily_snapshot.sql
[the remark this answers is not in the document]
sri-byte: Got you, Thanks @ifoukarakis

transform/mattermost-analytics/models/marts/web_app/platform/eg_performance_events.sql
```diff
@@ -0,0 +1,29 @@
+{{
```
ifoukarakis, line 1: `_platform__models.yml` with the model definitions and any tests is missing.

sri-byte: Added this file, no tests as of now. Want to get a copy of this PR up and running to work with the performance team and do a 2nd revision based on what we see

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
sri-byte: https://mattermost.atlassian.net/browse/MM-50797

transform/mattermost-analytics/models/marts/web_app/platform/eg_performance_events.sql
```diff
@@ -0,0 +1,29 @@
...
+
+WITH performance_rc AS (
+    SELECT 
+        'int_mm_telemetry_rc__performance_events' AS _source_relation,
+        {{dbt_utils.star(ref('int_mm_telemetry_rc__performance_events'))}}
+     FROM
+       {{ ref('int_mm_telemetry_rc__performance_events') }}
+{% if is_incremental() %}
+    WHERE received_at_date >= (SELECT MAX(received_at_date) FROM {{ this }}) 
+    WHERE _source_relation = 'int_mm_telemetry_rc__performance_events'
```
ifoukarakis, line 15: The `)` is at the wrong place, adding two `WHERE` in the parent statement.

transform/mattermost-analytics/models/marts/web_app/platform/eg_performance_events.sql
```diff
@@ -0,0 +1,29 @@
...
+{% endif %}
+), performance_prod AS (
+    SELECT
+        'int_mm_telemetry_prod__performance_events' AS _source_relation,
+        {{dbt_utils.star(ref('int_mm_telemetry_prod__performance_events'))}}
+     FROM
+       {{ ref('int_mm_telemetry_prod__performance_events') }}
+{% if is_incremental() %}
+    WHERE received_at_date >= (SELECT MAX(received_at_date) FROM {{ this }}) 
+    WHERE _source_relation = 'int_mm_telemetry_prod__performance_events'
```
ifoukarakis, line 25: Same as above. Also missing `endif`.

ifoukarakis (commented):
Added two comments about syntax issues.

transform/mattermost-analytics/models/marts/web_app/platform/eg_performance_events.sql
sri-byte: Fixed this and the dbt compile was successful

transform/mattermost-analytics/models/marts/web_app/platform/eg_performance_events.sql
sri-byte: Fixed this too

catalintomai:
Don't know if you plan to add an exposure like https://github.com/mattermost/mattermost-data-warehouse/commit/a607325276793190f678226759ee6605b35f621e and also, don't know if there ever were performance events captured by Segment (the work here: https://github.com/mattermost/mattermost-data-warehouse/pull/867/files#diff-949dd1e3900d7ccf08b3158a63dffe9ba66bc0fd4ca7d72eb42456a19f815192 only sources mm_telemetry_prod), but it'd be worth mentioning somewhere that these are Rudderstack only events.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
catalintomai: @sri-byte, are we sure that MM-50797 is the cleanup ticket Yiannis was referring to?

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_prod_performance_events.sql
```diff
@@ -0,0 +1,13 @@
+{{
+    config({
+        "materialized": "incremental",
+        "cluster_by": ['received_at_date'],
```
ifoukarakis, line 4: Looking at how long it takes to run on `TRANSFORM_XS`. Thinking it makes sense to run this on `TRANSFORM_L`.

sri-byte: Yes, I'm also seeing this

catalintomai: Can you share the query id? Thanks.

transform/mattermost-analytics/models/marts/web_app/platform/eg_performance_events.sql
```diff
@@ -0,0 +1,32 @@
...
+), 
+performance_prod AS (
+    SELECT
+        'int_mm_telemetry_prod_performance_events' AS _source_relation,
+        {{dbt_utils.star(ref('int_mm_telemetry_prod_performance_events'))}}
+     FROM
+       {{ ref('int_mm_telemetry_prod_performance_events') }}
+{% if is_incremental() %}
+    WHERE received_at_date >= (SELECT MAX(received_at_date) FROM {{ this }}) 
+    AND _source_relation = 'int_mm_telemetry_prod_performance_events'
```
ifoukarakis, line 27: Same issue with `_source_relation` here.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
sri-byte: I will remove the fields in the next iteration, after working with Harrison.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
sri-byte: Same as above, once I review the data with Harrison, we'll revisit this and make improvements.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__models.yml
sri-byte: I can think of one test, since they use the field `duration` to measure performance, I'll check with the team and add them.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__performance_events.sql
sri-byte: We probably don't need these, The next step for me is to clean the LookML view code, and then remove these fields from here.

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_prod_performance_events.sql
sri-byte: I shared the ID with Catalin, marking this as resolved.

transform/mattermost-analytics/models/marts/web_app/platform/eg_performance_events.sql
sri-byte: Fixed this.

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_rc_performance_events.sql
```diff
@@ -0,0 +1,13 @@
...
+        "materialized": "incremental",
+        "cluster_by": ['received_at_date'],
+    })
+}}
+
+SELECT * FROM 
+{{ ref('stg_mm_telemetry_rc__performance_events') }} 
+
+{% if is_incremental() %}
+    WHERE received_at_date >= (SELECT MAX(received_at_date) FROM {{ this }})
```
ifoukarakis, line 12: Same here. `received_at` seems like a better candidate.

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
```diff
@@ -0,0 +1,74 @@
+version: 2
+
+models:
+  - name: eg_performance_events
+    description: |
+      Events Group performance_events table containing metrics 
+      from web-app and web-app RC telemetry for category = performance.
+
+    columns:
+      - name: _source_relation
```
ifoukarakis, line 10: Tests are missing.

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_prod_performance_events.sql
```diff
@@ -0,0 +1,14 @@
...
+        "snowflake_warehouse": "transform_l"
+    })
+}}
+
+SELECT * FROM 
+{{ ref('stg_mm_telemetry_prod__performance_events') }}
+
+{% if is_incremental() %}
+    WHERE received_at_date >= (SELECT MAX(received_at_date) FROM {{ this }})
```
ifoukarakis, line 13: Incremental strategy should rely on `received_at`.

transform/mattermost-analytics/models/marts/web_app/platform/eg_performance_events.sql
```diff
@@ -0,0 +1,35 @@
...
+}}
+
+WITH performance_rc AS (
+    SELECT 
+        'int_mm_telemetry_rc_performance_events' AS _source_relation,
+        {{dbt_utils.star(ref('int_mm_telemetry_rc_performance_events'))}}
+     FROM
+       {{ ref('int_mm_telemetry_rc_performance_events') }}
+{% if is_incremental() %}
+    WHERE received_at_date >= (
```
ifoukarakis, line 15: Incremental strategy should rely on `received_at`.

ifoukarakis (changes requested):
Reviewed again with a bit more clear mind. Incremental logic has some challenges to address.

[Default incremental strategy](https://docs.getdbt.com/docs/build/incremental-models#about-incremental_strategy) is `merge`. This strategy requires a unique ID, which is not defined in any of the models.

A different approach would be to take advantage of the fact that `received_at` is always increasing and combine it with `append` strategy.  `append` strategy is also faster, as it does not require any changes to existing data.

Elaborating a bit more. New rows added after the last model execution will always have `received_at` values `>` (not `>=`) than the rows already added. This isn't true for `received_at_date`, as multiple executions of the model within the same day will result in same `received_at_date`. 

Note that making the switch to the `received_at` timestamp will require `>=` to be replaced with `>`.

Please also add tests.

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_rc_performance_events.sql
sri-byte: what would the cluster key be in this case?

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
sri-byte: I will have this ready by tomorrow morning. Will be adding some basic ones to begin

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_rc_performance_events.sql
ifoukarakis: Nice question. The clustering key may remain as is for int models. Since `received_at_date` is calculated from `received_at`, the query optimizer might take advantage of it. However we'll have to try it to see it.

It is also possible to take advantage of it to prune the data to be added using the `received_at_date`, and then implement the `append` strategy using `received_at`. But this is more challenging to implement and IMHO a premature optimization. Let's monitor incremental logic performance and implement it on a second stage if needed.

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_prod_performance_events.sql
sri-byte: Updated to `received_at`

transform/mattermost-analytics/models/marts/web_app/platform/eg_performance_events.sql
sri-byte: Updated to `received_at`

ifoukarakis: Both ends of the expression in the `WHERE` clause must use the same column (`received_at`).

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_rc_performance_events.sql
sri-byte: Updated the strategy to append using `received_at`

transform/mattermost-analytics/models/marts/web_app/platform/eg_performance_events.sql
sri-byte: Updated both clauses.

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
```diff
@@ -11,8 +11,11 @@ models:
       - name: context_page_referrer          
       - name: channel                        
       - name: context_app_namespace          
-      - name: id                             
-      - name: user_actual_id                 
+      - name: id
```
catalintomai, line 14: wonder if we can order/group these columns somehow (either alphabetically) or by importance (e.g. id first) or relationships (e.g. all context_* together, all date/time-related ones together, etc.)

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: context_screen_density         
+      - name: context_page_path              
+      - name: context_os_version             
+      - name: category                       
+      - name: duration                     
+      - name: num_of_request                 
+      - name: max_api_resource_size          
+      - name: longest_api_resource_duration  
+      - name: user_id                        
+      - name: userid
```
ifoukarakis, line 40: Looks like a typo in the implementation. Perhaps a `coalesce` between these two can simplify things.

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
```diff
@@ -0,0 +1,90 @@
+version: 2
+
+models:
+  - name: eg_performance_events
+    description: |
+      Events Group performance_events table containing metrics 
+      from web-app and web-app RC telemetry for category = performance.
+
+    columns:
+      - name: _source_relation
```
ifoukarakis, line 10: Since this column has specific values, an `accepted_values` test can be added.

ifoukarakis (commented):
Thanks for getting the test

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
```diff
@@ -0,0 +1,90 @@
...
+          combination_of_columns:
+            - timestamp
+            - received_at
+            - original_timestamp
+            - user_id
+            - user_actual_id
+            - category
+            - type
+            - duration
+            - id
```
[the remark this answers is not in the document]
sri-byte: this test, unique and not null test on `id` is failing, I tried this initially

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
sri-byte: I removed the field, there is no data in the field, I will remove the field in Looker as well, after this is merged. There are some changes I need to make to the Looker view.

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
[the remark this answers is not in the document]
sri-byte: I have tried multiple variations of unique combination of test , finally was able to crack it. The individual tests are giving me a tough time, since there is a lot of data that doesn't meet the criteria.

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_prod_performance_events.sql
```diff
@@ -0,0 +1,14 @@
+{{
+    config({
+        "materialized": "incremental",
+        "cluster_by": ['received_at_date'],
+        "snowflake_warehouse": "transform_l"
```
catalintomai, line 5: @sri-byte , do we have to set this here, specifically for this model? Or should we set it at a higher level for the whole set of performance models?

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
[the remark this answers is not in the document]
ifoukarakis: Unique failure on ID is strange. Means there might be duplicate data, so other steps need to be taken in order to deduplicate them.

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_prod_performance_events.sql
sri-byte: I don't think we need this for RC model based on the volume of data, and another point is we don't need this long term, since the incremental strategy for this is `append`. I plan to remove this once the tables are created.

ifoukarakis: I was just checking this and it's taking a long time. The default is `transform_xs`, so it makes sense to override it for this slow model. The query is rather simple, ~80% of the time is sorting.

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
```diff
@@ -0,0 +1,95 @@
...
+        tests:
+          - accepted_values:
+              values: ['int_mm_telemetry_rc_performance_events', 'int_mm_telemetry_prod_performance_events']
+      - name: context_page_referrer          
+      - name: channel                        
+      - name: context_app_namespace          
+      - name: id
+        tests:
+          - unique
+          - not null
```
ifoukarakis, line 20:
```suggestion
          - not_null
```

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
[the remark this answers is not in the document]
sri-byte: Failure on `id` is resolved, it was caused by duplicates from `received_at_date`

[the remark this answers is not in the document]
ifoukarakis: That's awesome! If `id` check is resolved, let's remove the `combination_of_columns` test. The uniqueness will always be true because an already unique column (`id`) is included in the columns. It is also taking a long time, make it impractical to be run on every PR/hourly.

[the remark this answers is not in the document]
sri-byte: Removed the `combination_of_columns` test.

ifoukarakis (approved):
I'm ok with the changes. Let's create a separate ticket for reviewing the columns in order to decide which ones to keep. Please paste the ticket link here for reference.

sri-byte:
Ticket to review the columns https://mattermost.atlassian.net/browse/MM-51427

PR #1183, opened 2023-02-28
MM-50553: alert on new table detection

ifoukarakis:
#### Summary

- [x] Add daily DAG that checks for new rudderstack event tables in a (configurable) list of schemas.
- [x] Update `rudder list` command to return non-zero exit code so that `KubernetesPodOperator` fails if command doesn't run.
- [x] Basic tests for DAG.
- [x] Add generic test for all DAGs that performs basic validation (DAG file can be loaded and DAG doesn't contain cycles).
- [x] Added a fix for CI/CD that was using wrong `requirements.txt` for airflow dependencies.

### DIFF
--- a/dags/monitoring/events.py
+++ b/dags/monitoring/events.py
@@ -16,6 +16,21 @@ from dags.kube_secrets import (
 )
 from plugins.operators.mattermost_operator import MattermostOperator
 
+doc_md = """
+### Nightly DBT dag
+
+#### Purpose
+
+This DAG triggers nightly tasks that check for new tables created by Rudderstack.
+
+The following options can be configured using Airflow variables:
+
+- `rudder_schemas` (JSON array of strings) - list of schemas to check.
+- `rudder_max_age` (integer) - include tables created in the past number of days defined by this variable.
+
+"""
+
+
 # Default arguments for the DAG
 default_args = {
     "depends_on_past": False,
@@ -34,6 +49,7 @@ dag = DAG(
     schedule_interval="0 12 * * *",
     catchup=False,
     max_active_runs=1,  # Don't allow multiple concurrent dag executions
+    doc_md=doc_md,
 )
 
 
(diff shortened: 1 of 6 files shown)


dags/tests/test_dag_integrity.py
```diff
@@ -0,0 +1,36 @@
...
+)
+
+DAG_PATHS = ALL_PATHS - EXCLUDED_PATHS
+
+
+@pytest.mark.parametrize("dag_path", [pytest.param(path, id=path.name) for path in DAG_PATHS])
+def test_dag_integrity(dag_path):
+    """Import DAG files and check for a valid DAG instance."""
+    dag_name = dag_path.name
+    module = _import_file(dag_name, dag_path)
```
catalintomai, line 21: 0/5, but since you don't use `dag_name` anywhere else, you could've extracted it inside `_import_file`.

catalintomai:
Would help fill in the context by also including the link to the Jira ticket.

ifoukarakis:
> Would help fill in the context by also including the link to the Jira ticket.

The ticket number's is in the PR's title. I can add the link as well if you prefer.

dags/monitoring/events.py
```diff
@@ -0,0 +1,80 @@
...
+    SNOWFLAKE_PASSWORD,
+    SNOWFLAKE_USER,
+)
+from plugins.operators.mattermost_operator import MattermostOperator
+
+# Default arguments for the DAG
+default_args = {
+    "depends_on_past": False,
+    "owner": "airflow",
+    "on_failure_callback": send_alert,
```
catalintomai, line 23: where is the alert being sent?

catalintomai (approved):

dags/monitoring/events.py
ifoukarakis: On the channel that has been configured. There's some space for improvement on the semantics of the `MattermostOperator`, but let's do it in a separate PR.

catalintomai:
that is the format/template used for the PRs of the other teams -see `mattermost-server` PRs for example. Guess it is easier to click on the link.

ifoukarakis:
> that is the format/template used for the PRs of the other teams -see `mattermost-server` PRs for example. Guess it is easier to click on the link.

Will do in the upcoming PRs.

PR #1205, opened 2023-03-13
Hotfix: XCOM return object instead of array

ifoukarakis:
#### Summary

JSON in file `/airflow/xcom/return.json` needs to be an object to map to key/value

#### Ticket Link

### DIFF
--- a/dags/monitoring/events.py
+++ b/dags/monitoring/events.py
@@ -60,7 +60,7 @@ def table_formatter(task_id, size=10):
     def format_tables(**kwargs):
         """Format result as table"""
         ti = kwargs['ti']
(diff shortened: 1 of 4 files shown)


munish7771 (approved):
looks good!

dags/monitoring/events.py
```diff
@@ -60,7 +60,7 @@ def table_formatter(task_id, size=10):
     def format_tables(**kwargs):
         """Format result as table"""
         ti = kwargs['ti']
-        result = ti.xcom_pull(task_ids=task_id)
+        result = ti.xcom_pull(task_ids=task_id)['new_tables']
```
munish7771, line 63: could we catch runtime ex in case xcom is not available ?

PR #1206, opened 2023-03-13
MM-51001: ignore exit code

ifoukarakis:
#### Summary

According to [airflow's documentation](https://airflow.apache.org/docs/apache-airflow-providers-cncf-kubernetes/stable/operators.html#how-does-xcom-work) XCOMs will be pushed only for tasks marked as `State.SUCCESS`. Current process is exiting with exit code > 0 if new tables are detected. This PR addresses this issue.

- [x] Ignore exit code of process.
- [x] Replace trigger rule with `ShortCircuitOperator` that checks if the list of tables is non-empty.

#### Ticket Link

### DIFF
--- a/dags/monitoring/events.py
+++ b/dags/monitoring/events.py
@@ -60,8 +60,7 @@ def short_circuit_on_no_new_tables(task_id):
         """Short circuit if no new tables"""
         ti = kwargs['ti']
         xcom_result = ti.xcom_pull(task_ids=task_id)
-        is_xcom_none_or_empty = not xcom_result or not xcom_result.get('new_tables', [])
-        return not is_xcom_none_or_empty
+        return bool(xcom_result and xcom_result.get('new_tables', []))
 
     return _do_short_circuit
 
--- a/dags/monitoring/events.py
+++ b/dags/monitoring/events.py
@@ -3,8 +3,7 @@ from datetime import datetime, timedelta
 from airflow import DAG
 from airflow.contrib.operators.kubernetes_pod_operator import KubernetesPodOperator
 from airflow.models import Variable
-from airflow.operators.python_operator import PythonOperator
-from airflow.utils.trigger_rule import TriggerRule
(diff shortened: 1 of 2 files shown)


dags/monitoring/events.py
```diff
@@ -56,11 +55,22 @@
 )
 
 
+def short_circuit_on_no_new_tables(task_id):
+    def _do_short_circuit(**kwargs):
+        """Short circuit if no new tables"""
+        ti = kwargs['ti']
+        xcom_result = ti.xcom_pull(task_ids=task_id)
+        is_xcom_none_or_empty = not xcom_result or not xcom_result.get('new_tables', [])
+        return not is_xcom_none_or_empty
```
munish7771, line 64: return not xcom_result and not xcom_result.get('new_tables', [])

munish7771 (commented):
are xcoms getting cleared from airflow db after dag finishes?

dags/monitoring/events.py
ifoukarakis: > return not xcom_result and not xcom_result.get('new_tables', [])

There's an extra `not` in line 64 that will be missed this way.

ifoukarakis:
> are xcoms getting cleared from airflow db after dag finishes?

`clean_xcom` is applied after the dag finishes.

dags/monitoring/events.py
munish7771: it should not 
not (not a or not b) == not a and not b

munish7771:
> > are xcoms getting cleared from airflow db after dag finishes?
> 
> `clean_xcom` is applied after the dag finishes.

is there a task or is airflow doing it internally?
I faced an issue with xcom filling up with this version of airflow even after dag finishing. Had to create a task for that.

ifoukarakis:
> 

AFAIK it should be done as part of the DAG or have a separate DAG that cleans up XCOM.

munish7771 (approved):

PR #1210, opened 2023-03-15
MM-51105: cleanup DAG for event table monitoring

ifoukarakis:
#### Summary

Fixes/improvements related to the DAG reporting new event tables.

- [x] Properly define `attachments` as a template field in `MattermostOperator`. This allows for proper variable replacement.
- [x] Make default `text` value equal to `null` in `MattermostOperator`. This field can be null if `attachments` is defined.
- [x] Convert message to list (from table). The formatting was getting messed if the number of tables is small.
- [x] Cleanup xcom in the beginning. This is a limitation of current version of airflow used internally. If a task is skipped, downstream tasks are skipped, even if a proper `trigger_rule` has been set. This was a [known issue](https://github.com/apache/airflow/issues/7858), fixed in recent versions of airflow. The main difference now is that the DAG will always start with the XCOM cleanup, leaving XCOM values available. This can actually be useful for debugging.

#### Ticket Link

### DIFF
--- a/dags/monitoring/events.py
+++ b/dags/monitoring/events.py
@@ -4,9 +4,7 @@ from airflow import DAG
 from airflow.contrib.operators.kubernetes_pod_operator import KubernetesPodOperator
 from airflow.models import Variable
 from airflow.operators.python_operator import PythonOperator, ShortCircuitOperator
-from tabulate import tabulate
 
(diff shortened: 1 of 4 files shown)


ifoukarakis:
Closed due to incorrect branch name.

PR #1214, opened 2023-03-20
MM - 51427 Removing null columns from performance_events

sri-byte:
Impact: Removes null value columns from eg_performance_events. Columns were identified using dbt profiler.

https://mattermost.atlassian.net/browse/MM-51427

### DIFF
--- a/transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_prod_performance_events.sql
+++ b/transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_prod_performance_events.sql
@@ -6,7 +6,7 @@
     })
 }}
 
-SELECT * FROM 
+SELECT {{ dbt_utils.star(ref('stg_mm_telemetry_prod__performance_events')) }} FROM 
 {{ ref('stg_mm_telemetry_prod__performance_events') }}
 
 {% if is_incremental() %}
--- a/transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_rc_performance_events.sql
+++ b/transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_rc_performance_events.sql
@@ -5,7 +5,7 @@
     })
 }}
 
-SELECT * FROM 
+SELECT {{ dbt_utils.star(ref('stg_mm_telemetry_rc__performance_events')) }} FROM 
 {{ ref('stg_mm_telemetry_rc__performance_events') }} 
 
 {% if is_incremental() %}


sri-byte:
When this is approved, I will make the Looker PR and ready the script to drop the columns, merge the Looker PR, then drop columns and merge this PR

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
```diff
@@ -11,71 +11,48 @@ models:
...
-      - name: received_at
-      - name: event_date
+      -name: anonymous_id
+    	-name: received_at
+    	-name: sent_at
+    	-name: original_timestamp
+    	-name: timestamp
+    	-name: context_ip
+    	-name: event
+    	-name: event_text
```
catalintomai, line 25: `event` and `event_text` look to be new columns - are they needed in looker? (did we miss them in the initial iteration)?

sri-byte: They aren't in Looker, they are the default rudderstack columns, we get these from the macro https://github.com/mattermost/mattermost-data-warehouse/blob/master/transform/mattermost-analytics/macros/rudderstack/get_rudderstack_columns.sql

sri-byte:
> One q that comes to mind - are they always going to be null; are they null now because of an upstream bug or because of other reasons - like code removal, etc.

Not sure if they're always going to be null, we should check that with a dev and see how that works. It has been null up until now and also not used in Looker.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__models.yml
```diff
@@ -28,67 +28,43 @@ models:
...
-      - name: sort                           
-      - name: team_id                        
-      - name: version                        
-      - name: keyword                        
-      - name: count                          
-      - name: gfyid                          
-      - name: field                          
-      - name: plugin_id                      
-      - name: installed_version              
-      - name: group_constrained
```
catalintomai, line 71: `group_constrained` looks to be used by looker, so by the description above "using only columns needed by looker." - should be kept?

sri-byte: sorry, it exists in LookML, but the field is `null`, it will be removed from the LookML in the following PR, its not used in performance_events dashboard.

sri-byte: no, this is exclusively for `performance_events` table. This table is created with the filter `where category = 'performance`` in the below models, the profiler was applied on the final performance_events table, so this column can be dropped in this particular case.

transform/mattermost-analytics/models/marts/web_app/platform/_platform__models.yml
catalintomai: Wondering what value do we get by getting the `event` column for performance events. The only value it contains is `event`.

sri-byte: its only 1 value `event` from this column

catalintomai: yes..

sri-byte: We'll scrub them in the next iterations, now that we've removed the `null`

transform/mattermost-analytics/models/staging/mm_telemetry_rc/stg_mm_telemetry_rc__performance_events.sql
```diff
@@ -1,15 +1,9 @@
...
+, "context_useragent", "context_app_name", "context_locale", "context_screen_density" 
 , "context_page_path", "context_os_version", "category" , "duration" , "num_of_request", "max_api_resource_size"
-, "longest_api_resource_duration" , "user_id", "root_id", "post_id", "sort" 
-, "team_id", "keyword", "count", "gfyid", "field"
-, "plugin_id", "installed_version", "group_constrained", "value", "include_deleted" 
-, "role", "privacy", "scheme_id", "metric", "error", "num_invitations_sent" 
-, "num_invitations", "channel_sidebar", "app", "method", "remaining", "screen" 
-, "filter", "version", "uuid_ts"] 
--%}
+, "longest_api_resource_duration" , "user_id", "count", "uuid_ts"] -%}
```
ifoukarakis, line 6: `uuid_ts` can also be removed as it's used internally by Rudderstack.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__models.yml
```diff
@@ -28,67 +28,42 @@ models:
       Reconstructed `performance_events` table from schema `MM_TELEMETRY_PROD` using only columns needed by looker.
```
catalintomai, line 28: I think we should also mention that the remaining columns have non-null values, as dictated by running the profiler at date X (e.g. March 21 2023) - at least until we have a regular profiler running job.

catalintomai:
@sri-byte , it looks to me like there is an underlying assumption here, which is that the columns being removed will continue to be null in the future as well - which is likely, but not guaranteed. Is there a way to monitor if one of the removed columns suddenly starts having valid values (e.g. maybe a bug has been fixed) - prob. as a result of running the profiler as part of some job. At the minimum we should open a ticket to address this in the future.

transform/mattermost-analytics/models/staging/mm_telemetry_rc/stg_mm_telemetry_rc__performance_events.sql
```diff
@@ -1,15 +1,9 @@
 {%- set include_columns = ["context_page_referrer", "channel", "context_app_namespace", "user_actual_id"
 , "context_library_name", "type", "context_app_version" , "user_actual_role", "context_page_url", "context_os_name"
```
ifoukarakis, line 2: The following columns can also be removed: `context_page_path`, `context_os_name`, `context_os_version`, `context_page_referrer`, `context_page_search`, `context_page_title`, `context_page_url`.

They contain < 10 non null values and they are all junk.

ifoukarakis (changes requested):
Added a couple of suggestions. Perhaps it also makes sense to include renaming `eg_performance_events` to `grp_performance_events` as part of this PR.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__models.yml
sri-byte: The dashboard uses none of these fields, even if they were non-null in the future, it wouldn't apply to this use case.

transform/mattermost-analytics/models/staging/mm_telemetry_rc/stg_mm_telemetry_rc__performance_events.sql
sri-byte: Removed these

transform/mattermost-analytics/models/staging/mm_telemetry_rc/stg_mm_telemetry_rc__performance_events.sql
sri-byte: Removed `uuid_ts`

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__models.yml
catalintomai: Does that mean that we remove columns that have non-null values as well? Or was the initial reason for removing them the fact that they were null? The title of the PR says that we remove the null columns, but the comment here say that we remove them because they are not referenced in Looker dashboards

ifoukarakis: There's also some columns that have up to 8 non-null values (out of the 1+billion rows). These values are junk, making these columns practically always null.

ifoukarakis (approved):
🚀

PR #1216, opened 2023-03-20
MM-51305: draft model for release quality

ifoukarakis:
#### Summary

Draft model for release quality. 
- [x] Create staging models to extract information from JIRA issues.
- [x] Evaluate data to see if it matches documented model and identify gaps.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51305

### DIFF
--- a/transform/mattermost-analytics/models/marts/release/dim_fix_versions.sql
+++ b/transform/mattermost-analytics/models/marts/release/dim_fix_versions.sql
@@ -5,10 +5,9 @@ WITH unparsed_fix_versions AS (
         issue_id,
         value:name::string as fix_version_name,
         -- Break down different variations of target version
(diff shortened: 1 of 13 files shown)


transform/mattermost-analytics/models/marts/release/dim_fix_versions.sql
```diff
@@ -0,0 +1,36 @@
+
+WITH unparsed_fix_versions AS (
```
catalintomai, line 2: should we also handle smth. like `triage` in the fix versions names set?

ifoukarakis: Following discussion with @amyblais, only fix versions with a version in them (vX.Y) have some value for the current use case. Additional conventions in "fix version" field can be added in the future, if needed.

catalintomai (approved):

PR #1224, opened 2023-03-29
MM-51730: run mattermost analytics nightly

ifoukarakis:
#### Summary

- [x] Adjust models to run nightly. 
- [x] Tag slow tests with a separate tag. This will allow decoupling slow test/nightly execution in the future.
- [x] Update nightly DAG to trigger nightly job.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51730

### DIFF
--- a/dags/transformation/dbt_nightly.py
+++ b/dags/transformation/dbt_nightly.py
@@ -73,5 +73,25 @@ dbt_run_cloud_nightly = KubernetesPodOperator(
     dag=dag,
 )
 
+dbt_run_cloud_mattermost_analytics_nightly = KubernetesPodOperator(
+    **pod_defaults,
+    image=MATTERMOST_DATAWAREHOUSE_IMAGE,  # Uses latest build from master
+    task_id="dbt-cloud-mattermost-analytics-nightly",
+    name="dbt-cloud-mattermost-analytics-nightly",
+    secrets=[
+        DBT_CLOUD_API_ACCOUNT_ID,
+        DBT_CLOUD_API_KEY,
+        SNOWFLAKE_ACCOUNT,
+        SNOWFLAKE_USER,
+        SNOWFLAKE_PASSWORD,
+        SNOWFLAKE_TRANSFORM_ROLE,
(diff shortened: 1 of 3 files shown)


dags/transformation/dbt_nightly.py
```diff
@@ -73,5 +73,25 @@
...
+        SNOWFLAKE_TRANSFORM_SCHEMA,
+        SSH_KEY,
+    ],
+    env_vars=env_vars,
+    arguments=["python -m  utils.run_dbt_cloud_job 254981 \"Mattermost Analytics DBT nightly\""],
+    dag=dag,
+)
 
-dbt_run_cloud_nightly
+dbt_run_cloud_mattermost_analytics_nightly >> dbt_run_cloud_nightly
```
munish7771, line 97: do we need this dependency?
cloud nightly should run irrespective of fail/success of analytics nightly.

catalintomai (approved):

munish7771 (changes requested):

dags/transformation/dbt_nightly.py
ifoukarakis: The reason they run consequently is to spread the load on the database over time. If they run in parallel, there will be a lot of load in some of the already heavily utilized sources. Order can change.

Another approach would be a different DAG with a different schedule. This however wouldn't ensure that the DAGs do not run in parallel.

munish7771: In case the upstream fails, cloud nightly will not be triggered.

ifoukarakis: That's true. Shall I just add a trigger rule on `dbt_run_cloud_nightly` so that it always run?

munish7771: yes, looks good.

ifoukarakis: Adjusted mattermost-analytics to run second, switched tigger rule to `ALL_DONE` (from default `ALL_SUCCESS`).

munish7771 (approved):
lgtm!

PR #1227, opened 2023-04-06
MM-51085: sync additional fields from updated in product trial activation from to SFDC

ifoukarakis:
#### Summary

- [x] Source for trial requests.
- [x] Validation of source data.
- [x] Staging layer for trial requests.
- [x] Setup model(s) to be used for syncing with Salesforce. 
- [x] Seed data for valid salesforce country names.
- [x] Macro for validating emails.
- [x] Tests on mart layer.

Tests are currently failing due to a bug in ingestion of input data resulting in duplicate data. As soon as it's fixed, I'll re-run the tests.

I'd love feedback in the following decisions:
- Invalid emails are not synced to SF. There's a test on source data that has a soft-limit on invalid emails. We can discuss with other teams on whether this should be expected or not.
- First 40 characters of email are used in case `name` is empty string or null. An alternative would be to use a fixed string (i.e. `Unknown`).
 - There is no intermediate layer or incremental logic as (a) the data are not reused elsewhere, (b) number of rows is too low.

An exposure will be added after this PR is merged and the hightouch syncs have been created.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51085


transform/mattermost-analytics/models/marts/sales/hightouch/fct_in_product_trial_requests.sql
```diff
@@ -0,0 +1,48 @@
...
+        end as company_size,                                -- Mapping lower threshold
+        company_name,                                       -- Mapped to field company of lead
+        -- Salesforce lowercases email
+        lower(coalesce(contact_email, email)) as email,     -- Mapped to field email of lead
+        case
+           -- Fix inconsistency in country name so that it matches values expected by SF
+           when country_name = 'United States of America' then 'United States'
+           else country_name
+        end as country_name,                               -- Mapped to field country
+        start_at as trial_start_at,                         -- Mapped to request_a_trial_date__c and TBD
```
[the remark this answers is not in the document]
ifoukarakis: In order to build the hightouch sync, the model needs to be merged to master and run at least once. As soon as this is ready, an exposure can be added to provide all this information.

transform/mattermost-analytics/models/marts/sales/hightouch/fct_in_product_trial_requests.sql
```diff
@@ -0,0 +1,48 @@
...
+        num_users                                           -- Mapped to field numberofemployees of lead
+    from
+        {{ ref('stg_cws__trial_requests')}}
+)
+select
```
catalintomai, line 39: According to Kevin's latest comments - https://mattermost.atlassian.net/browse/MM-51085?focusedCommentId=149587 - looks like we need to create a campaign member as well(?) Also, according to John's description - "a lead entry should be created in SFDC if one does not already exist" - is that done by a different model?

ifoukarakis: There's a few things to clarify in today's meeting. I'll add them in the ticket. 

Regarding new vs existing items, a single model with a flag on whether a lead with the same email exists should be enough. Hightouch allows for updates based on models or queries. My idea was to have a single model and use the flag on hightouch to decide on whether an update or insert should happen.

catalintomai (approved):

munish7771 (changes requested):
can we also join with `campaignmember` in mart layer to know which records to insert to salesforce.

munish7771 (approved):
lgtm!

PR #1238, opened 2023-04-11
MM-51951: fix country name

ifoukarakis:
#### Summary

- [x] Add seed data for country code -> country name mapping
- [x] Normalize country name field from blapi. The column contains both country names and country codes. The updated script attempts to convert from country code to country name. If it fails, it keeps the original value.
- [x] Improve country code consistency (less nulls to be expected). 
- [x] State and state code should also be more consistent after this PR is merged.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51951

### DIFF
--- a/transform/snowflake-dbt/models/blapi/customers_with_onprem_subs.sql
+++ b/transform/snowflake-dbt/models/blapi/customers_with_onprem_subs.sql
@@ -39,7 +39,7 @@ WITH payment_addresses AS (
         COALESCE(pa.country_code, cc.code) AS country_code
     FROM
         payment_addresses pa
-        LEFT JOIN {{ ref('country_codes') }} cc ON pa.country = cc.code
+        LEFT JOIN {{ ref('country_codes') }} cc ON pa.country_code = cc.code
     WHERE
         row_num = 1
 ), customers_with_onprem_subs AS (
--- a/transform/snowflake-dbt/models/blapi/customers_with_onprem_subs.sql
+++ b/transform/snowflake-dbt/models/blapi/customers_with_onprem_subs.sql
@@ -5,7 +5,7 @@
   })
 }}
 
-WITH latest_credit_card_address AS (
+WITH payment_addresses AS (
     SELECT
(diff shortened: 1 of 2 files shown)


munish7771 (approved):
looks good!

transform/snowflake-dbt/models/blapi/customers_with_onprem_subs.sql
```diff
@@ -24,6 +24,24 @@ WITH latest_credit_card_address AS (
...
+        pa.line2,
+        pa.postal_code,
+        pa.city,
+        pa.state,
+        COALESCE(cc.name, pa.country) AS country,
+        pa.state_code,
+        COALESCE(pa.country_code, cc.code) AS country_code
+    FROM
+        payment_addresses pa
+        LEFT JOIN {{ ref('country_codes') }} cc ON pa.country = cc.code
```
catalintomai, line 42: `pa.country = cc.code` -> `pa.country_code = cc.code` ?

catalintomai (approved):

transform/snowflake-dbt/models/blapi/customers_with_onprem_subs.sql
ifoukarakis: > `pa.country = cc.code` -> `pa.country_code = cc.code` ?

It's practically the same due to the `COALESCE` exception. `addresses` seem to contain country codes in `country` field. Whenever country code or country from postal code mapping is null, the country code is used (for both!). However updating as it's a bit more readable.

A side note is that CTE `latest_credit_card_address` is used in other models. It can also be simplified and become more consistent. Moving it to a separate model can help avoid repeating code, keep data consistent and simplify things. However this is quite some effort.

catalintomai: "addresses seem to contain country codes in country field" - think we should open a bug on BLAPI/Nick team to address this long-term (they might improve other validations as well while they are in the area)

PR #1239, opened 2023-04-11
Fixed incremental logic to use updated_at

sri-byte:
Impact: Fixed incremental logic to use updated_at, currently it uses created_at which is always going to be the same for the workspace. If there is a change to the subscription it will be reflected by updated_at field, which should be used in the incremental logic


munish7771 (approved):
Looks good, but you need to do a full refresh of this model once after this change is deployed.

catalintomai:
@sri-byte , is this the preliminary solution prior to removing the incremental approach? Or have you decided not to remove the incremental approach in the end?

PR #1243, opened 2023-04-13
Removes subscription_version_id_mod filter

sri-byte:
Impact: Removes subscription_version_id_mod filter from on_prem subscriptions view

### DIFF
--- a/transform/snowflake-dbt/models/blapi/onprem_subscriptions.sql
+++ b/transform/snowflake-dbt/models/blapi/onprem_subscriptions.sql
@@ -24,8 +24,7 @@ WITH latest_payment AS (
     FROM {{ source('blapi', 'subscriptions_version') }} s
     JOIN {{ source('blapi', 'products') }} p ON s.product_id = p.id
     JOIN latest_payment ON s.id = latest_payment.subscription_id AND latest_payment.row_num = 1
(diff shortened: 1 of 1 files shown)


transform/snowflake-dbt/models/blapi/onprem_subscriptions.sql
```diff
@@ -24,8 +24,7 @@ WITH latest_payment AS (
     FROM {{ source('blapi', 'subscriptions_version') }} s
     JOIN {{ source('blapi', 'products') }} p ON s.product_id = p.id
     JOIN latest_payment ON s.id = latest_payment.subscription_id AND latest_payment.row_num = 1
-    WHERE s.subscription_version_id_mod
-        AND p.name != 'Mattermost Cloud'
+    WHERE p.name != 'Mattermost Cloud'
```
munish7771, line 27: we can use p.name not in ()

munish7771 (approved):

catalintomai:
@sri-byte , can you share Nick's input on the issue?

catalintomai:
@sri-byte , please add more info wrt to "Removes subscription_version_id_mod filter" - why do we need/want to remove it, what ticket(s) are impacted, etc to the description (or to the Jira ticket and reference here).

PR #1247, opened 2023-04-19
MM-51943: fixes for contact us campaign lead inserts

ifoukarakis:
#### Summary

- [x] Add email validation. 
- [x] Stricter email validation as new cases were detected in the failing syncs.
- [x] Move email validation to a macro so that it's reusable and consistent.
- [x] Update logic for new records to check column `SFID` rather than `dwh_external_id__c` as the second may be null.
- [x] Try to insert only first record in case multiple contact form attempts from the same email exist. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51943


munish7771 (approved):
lgtm!

PR #1249, opened 2023-04-19
MM-51943: fix duplicate campaign member inserts on same batch

ifoukarakis:
#### Summary

Removes duplicate emails in the same batch. Keeps the row with the most recent request to contact time.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-51943

### DIFF
--- a/transform/snowflake-dbt/models/hightouch/blapi/contact_us_campaignmember_to_insert_contact.sql
+++ b/transform/snowflake-dbt/models/hightouch/blapi/contact_us_campaignmember_to_insert_contact.sql
@@ -7,5 +7,7 @@
 with campaignmembers_to_insert as (
     select * from {{ ref('contact_us_campaign_fact') }}
     where campaignmember_sfid is null and not lead_exists and contact_exists
+    -- For each insert batch, keep the most recent record according to "request to contact us"
+    qualify row_number() over (partition by email order by request_to_contact_us_date desc) = 1
 )
 select * from campaignmembers_to_insert


catalintomai:
@ifoukarakis , why "most recent request to contact time." as opposed to the oldest/first request to contact?

ifoukarakis:
> 

This sync updates a `status` field, so best guess here is that this is the one needed. 

Regarding leads I thought I saw an update in the queries but I cannot find a matching sync. Will check further and if I don't find something I'll also change it to pick the latest entry.

catalintomai:
> > 
> 
> @catalintomai The sync related to this PR updates a `status` field, so best guess here is that this is the one needed.
> 
Maybe it doesn't matter(?), as long we pick only one - my thinking was that the first one is the on that matters, the next ones being extra/spurious? Also, there are quite a few `with campaignmembers_to_insert`-like models, not clear if all of them have corresponding Hightouch syncs.

ifoukarakis:
> > > 
> > 
> > 
> > @catalintomai The sync related to this PR updates a `status` field, so best guess here is that this is the one needed.
> 
> Maybe it doesn't matter(?), as long we pick only one - my thinking was that the first one is the on that matters, the next ones being extra/spurious - `status` looks to be hardcoded? Also, there are quite a few `with campaignmembers_to_insert`-like models, not clear if all of them have corresponding Hightouch syncs.

Let's discuss with the team later. I also found that model `contact_us_campaign_with_contact_update` is not part of any sync.

ifoukarakis:
Closing as this was fixed by https://github.com/mattermost/mattermost-data-warehouse/pull/1253

PR #1255, opened 2023-04-20
Chore: Add stripe tables to mattermost-analytics

munish7771:
#### Summary

Added following staging tables for source `Stripe`. Required for adding `cloud enterprise trials` to salesforce.
1. Customers
2. Subscriptions
3. Products
#### Ticket Link

N/A


transform/mattermost-analytics/models/staging/stripe/stg_stripe__customers.sql
```diff
@@ -0,0 +1,50 @@
...
+        cards,
+        created,
+        currency,
+        delinquent,
+        email,
+        id,
+        invoice_prefix,
+        invoice_settings,
+        livemode,
+        metadata:"contactfirstname"::varchar as contactfirstname,
```
ifoukarakis, line 21:
```suggestion
        metadata:"contactfirstname"::varchar as contact_first_name,
```

transform/mattermost-analytics/models/staging/stripe/stg_stripe__customers.sql
```diff
@@ -0,0 +1,50 @@
...
+        metadata:"contactlastname"::varchar as contactlastname,
```
ifoukarakis, line 22:
```suggestion
        metadata:"contactlastname"::varchar as contact_last_name,
```

transform/mattermost-analytics/models/staging/stripe/stg_stripe__products.sql
```diff
@@ -0,0 +1,39 @@
...
+    select * from {{ source('stripe_raw', 'products') }}
+
+),
+
+products as (
+
+    select
+        active,
+        attributes,
+        created,
```
ifoukarakis, line 12: Timestamps should follow the postfix `_at`. 
```suggestion
        created AS created_at,
```

transform/mattermost-analytics/models/staging/stripe/stg_stripe__products.sql
```diff
@@ -0,0 +1,39 @@
...
+        livemode,
+        metadata:"cws-product-family"::varchar as cws_product_family,
+        metadata:"cws-sku-name"::varchar as cws_sku_name,
+        metadata:"cws-allowable-addons"::varchar as cws_allowable_addons,
+        metadata,
+        name,
+        object,
+        shippable,
+        type,
+        updated,
```
ifoukarakis, line 25: Same here
```suggestion
        updated as updated_at,
```

transform/mattermost-analytics/models/staging/stripe/stg_stripe__products.sql
```diff
@@ -0,0 +1,39 @@
...
+        attributes,
+        created,
+        description,
+        id,
+        images,
+        livemode,
+        metadata:"cws-product-family"::varchar as cws_product_family,
+        metadata:"cws-sku-name"::varchar as cws_sku_name,
+        metadata:"cws-allowable-addons"::varchar as cws_allowable_addons,
+        metadata,
```
ifoukarakis, line 20: Do we need the metadata? Better to extract everything here.

ifoukarakis (commented):
Let's update column names to match naming conventions.

transform/mattermost-analytics/models/staging/stripe/stg_stripe__products.sql
sri-byte: lets keep the metadata, we could extract everything, but if they add new fields or if they change the name of the JSON object which might lead to nulls because we need to update the fields, it will be easy to compare.

ifoukarakis: Keeping the metadata and allowing downstream models to use them can introduce ambiguity. A typical scenario would be a downstream model extracting a field from `metadata` already extracted in the current staging model and giving it a different name than the one in the current model. 

The staging layer is also useful for standardizing naming and structure. If new fields are added in the `metadata` json field,  any renaming and handling must happen to the staging model. This approach also allows introducing tests to detect issues exactly like the ones described by @sri-byte. For example, a `not_null` test can be added on the extracted fields if we know that all rows must have a specific key in `metadata`.

munish7771: Removing metadata and adding only the columns that we need right now. In case any new field is needed from metadata we can always update the staging models.

ifoukarakis (changes requested):
Tests are failing and there are some issues in the yaml files. 

Another thing I'm not sure about is whether stitch columns should be kept or not. If there's no known use for them, they can be removed. If in the future they are needed they can be re-added.

transform/mattermost-analytics/models/staging/stripe/stg_stripe__customers.sql
```diff
@@ -0,0 +1,51 @@
...
+        _sdc_table_version,
+        default_source,
+        shipping,
+        tax_ids,
+        address,
+        default_card,
+        discount,
+        updated_by_event_type,
+        updated as _updated_at,
+        created as _created_at
```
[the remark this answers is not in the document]
munish7771: these are for incremental loads.

transform/mattermost-analytics/models/staging/stripe/_stripe__models.yml
```diff
@@ -0,0 +1,226 @@
...
+      - default_payment_method:
+          description: "The default payment method for a customer's subscription."
+      - cancel_at:
+          description: "The date and time when a subscription will be canceled."
+      - pause_collection:
+          description: "A flag indicating whether collection of payment for a subscription has been paused."
+      - discount:
+          description: "The discount applied to a subscription."
+      - days_until_due:
+          description: "The number of days before a subscription is due for payment."
```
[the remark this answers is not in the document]
munish7771: fixed.

transform/mattermost-analytics/models/staging/stripe/stg_stripe__customers.sql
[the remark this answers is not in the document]
ifoukarakis: Is there any difference with `updated_at`?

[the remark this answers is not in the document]
munish7771: removed.

ifoukarakis (approved):

PR #1259, opened 2023-04-21
MM-52375 Changing cadence for hourly runs to nightly 

sri-byte:
Impact: Due to our usage/consumption patterns, we are changing the frequency of building some of the dbt models from hourly to nightly. This will help reduce unnecessary computation (and costs at the same time)


sri-byte:
@ifoukarakis I sat down with @catalintomai and we did a scrub of existing models and moved some of the hourly models to nightly. I merged the previous PR into this PR, since it was from a forked repo as I had access issues.

catalintomai (approved):

ifoukarakis (changes requested):
Had a quick look. The following models will run in both hourly and nightly jobs:

- `events.nightly.calls_events`
- `events.nightly.feature_flag_telemetry`
- `events.nightly.focalboard_event_telemetry`
- `events.nightly.mobile_events`
- `events.nightly.rudder_webapp_events`
- `events.nightly.segment_webapp_events`

Reason for this is that they are tagged with `preunion` tag. Moving them to nightly might require editing the DBT jobs.

I'd also like to check the timing of these models and any downstream dependencies, but this would take some time.

ifoukarakis:
Following up after checking the timings, here's a few more suggestions:
- `util.telemetry_tables` ~ 110s
- `util.telemetry_columns` ~ 250s

Another thing to consider is that most of the models moving to nightly are used for populating data on Salesforce. Given that all these updates run hourly, it might be possible that data will not be available in upstream models.

sri-byte:
Closing PR, as this interferes with some real time updates with SFDC.

PR #1262, opened 2023-04-27
MM-51572: fix customer <--> opportunity mapping

ifoukarakis:
#### Summary

Current logic "connects" customers with existing opportunities based on `subs_id__c`. However this column is not set via any sync in hightouch. Similar information is available in `dwh_external_id__c`. This column is now used to offer an alternative join.

Comparing the two models show that the new version matches 43 (out of 53) records that weren't matched before.

Merging this PR should resolve failures in blapi cloud opportunity sync.

#### Ticket Link

### DIFF
--- a/transform/snowflake-dbt/models/blapi/customers_with_cloud_subs.sql
+++ b/transform/snowflake-dbt/models/blapi/customers_with_cloud_subs.sql
@@ -123,17 +123,24 @@ WITH latest_credit_card_address AS (
     SELECT
         customers_with_cloud_subs.subscription_id,
         COALESCE(
-            opportunity.dwh_external_id__c,
+            o1.dwh_external_id__c,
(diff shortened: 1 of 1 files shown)


catalintomai:
@ifoukarakis , that is impressive! -  does that mean that we cannot avoid using (and populating) `dwh_external_id__c` by the models from the new`mattermost-analytics` repo?

ifoukarakis:
> @ifoukarakis , that is impressive! - does that mean though that we will not be able to avoid using (and populating) `dwh_external_id__c` by the models from the new`mattermost-analytics` repo?

For opportunities, `dwh_external_id__c` is a hash of `subscription_id` from blapi `subscription_version`. Some migration might need to happen if we want to move i.e. from blapi's id to stripe id.

munish7771 (approved):

PR #1266, opened 2023-04-28
MM-51752: do not update accounts without contacts

ifoukarakis:
#### Summary

Opening this PR to trigger discussion, as I'm 3/5 on this solution.

This PR fixes the related broken sync. The reason that the sync fails is because it tries to update the salesforce `Account` object, but it fails to find it. The respective account are not there because it is only created if a `Contact` object exists ([source](https://github.com/mattermost/mattermost-data-warehouse/blob/master/transform/snowflake-dbt/models/hightouch/freemium/professional_subscription/freemium_account.sql#L55)). I couldn't clarify why this needs to happen.

There are two possible solutions:
- Don't update accounts without contacts. This is the solution proposed on this PR.
- Create accounts even if there is no contact. Perhaps contact object can also be added if missing.

#### Ticket Link

### DIFF
--- a/transform/snowflake-dbt/models/hightouch/freemium/professional_subscription/freemium_account_update.sql
+++ b/transform/snowflake-dbt/models/hightouch/freemium/professional_subscription/freemium_account_update.sql
@@ -5,7 +5,20 @@
   })
 }}
 
-WITH customers_with_cloud_paid_subs as (
+WITH existing_contacts AS (
+    SELECT
+        contact.sfid,
+        contact.email,
+        contact.dwh_external_id__c,
+        contact.ownerid,
+        ROW_NUMBER() over (
+            PARTITION BY contact.email
+            ORDER BY
+                createddate DESC
+        ) AS row_num
+    FROM
+        {{ ref('contact') }}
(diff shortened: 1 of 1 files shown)


catalintomai:
@ifoukarakis , do you think we should run this by Kevin/John first?

ifoukarakis:
> @ifoukarakis , what about running this by Kevin/John first? cc: @kevin-fayle-mm , @johndavidlugtu

Yes, that's a good idea.

There's one more thing that seems a bit off in this model. It's the email domain - clearbit domain matching.

johndavidlugtu:
Kevin says:
1. Create account
2. Create contact
3. Associate contact with account

cc: @ifoukarakis took down notes from what Kevin said

catalintomai:
> Kevin says:
> 
> 1. Create account
> 2. Create contact
> 3. Associate contact with account
> 
> cc: @ifoukarakis took down notes from what Kevin said

For now @kevin-fayle-mm will be creating the contact manually in SFDC, if more of these cases occur, we'll revisit handling this automatically.

ifoukarakis:
Closing as the fix will be on the data.

PR #1265, opened 2023-04-27
Chore: Removing deprecated models (cloud free and trial requests)

munish7771:
#### Summary

Removing the models under as they will be deprecated after the new release. 
1. hightouch/freemium/cloud_starter/*
2. hightouch/freemium/etr/*

There is an exception for model `hightouch/freemium/cloud_starter/cs_signup_campaign_facts` which is used by a downstream model in `data quality` layer and is not removed.

#### Ticket Link

N/A
This PR is among the deployment steps for -> https://github.com/mattermost/mattermost-data-warehouse/pull/1265

#### Deployment steps
- [ ] Move the tables to archive with lifecycle of 30 days in snowflake.
- [x] Disable/Delete the downstream Hightouch syncs.

PR #1274, opened 2023-05-05
MM-51752: remove duplicate rows

ifoukarakis:
#### Summary

Remove join with Salesforce account as it is already happening on the parent upstream model. This was creating a lot of duplicate rows. Main reason was the matching on gmail account.

#### Ticket Link


catalintomai:
Shouldn't this model be removed as well as a follow-up to https://github.com/mattermost/mattermost-data-warehouse/pull/1265 ? cc: @munish7771

munish7771:
> Shouldn't this model be removed as well as a follow-up to #1265 ? cc: @munish7771

This is for professional opportunity. (Cloud starter and trial requests were deprecated)

catalintomai (approved):

PR #1276, opened 2023-05-11
MM-51752: fix empty account name for gmail users

ifoukarakis:
#### Summary

Use predefined string for account name when contact email is a gmail account.

Next step would be to update hightouch sync after the change has been merged and run.

#### Ticket Link

### DIFF
--- a/transform/snowflake-dbt/models/hightouch/blapi/blapi_cloud_account.sql
+++ b/transform/snowflake-dbt/models/hightouch/blapi/blapi_cloud_account.sql
@@ -8,7 +8,8 @@
 WITH cloud_accounts_to_sync as (
     SELECT
         customers_with_cloud_subs.*,
-        '0053p0000064nt8AAA' AS ownerid
+        '0053p0000064nt8AAA' AS ownerid,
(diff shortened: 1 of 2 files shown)


transform/snowflake-dbt/models/hightouch/blapi/blapi_cloud_account.sql
```diff
@@ -8,7 +8,8 @@
 WITH cloud_accounts_to_sync as (
     SELECT
         customers_with_cloud_subs.*,
-        '0053p0000064nt8AAA' AS ownerid
+        '0053p0000064nt8AAA' AS ownerid,
+        COALESCE(customers_with_cloud_subs.domain, '{{ var('unknown_account_name') }}' ) AS account_name
```
catalintomai, line 12: still have mixed feelings about this, but I guess it got the approval 
of its consumers.

catalintomai (approved):

transform/snowflake-dbt/models/hightouch/blapi/blapi_cloud_account.sql
ifoukarakis: Agree. We should spend some time to document the different conventions used, at least for the most common items.

PR #1278, opened 2023-05-19
MM-51752: Fix incorrect match of subscription to accounts

ifoukarakis:
#### Summary

The current model tries to locate an account either via external id match or domain match. This introduces the following errors:
- If multiple accounts with the same domain exist, then all of them are matched because of the join. 
- If the domain has not been set for the account, no match occurs.

The solution proposed in this PR is to make matching a bit more solid. To do so, the following priority is used for matching accounts:
- Account with same existing `dwh_external_id__c` (if any).
- Account from a contact matching the same email (if any).
- Account from a lead matching the same email (if any).
- Account from domain match (if any). 

In case of multiple matches due to:
- multiple accounts sharing the same domain or
- multiple contacts/leads sharing the same email,

then the items are sorted by the different matches, with the most matches appearing first. The first row is only picked.

#### Ticket Link

### DIFF
--- a/transform/snowflake-dbt/models/hightouch/freemium/professional_subscription/freemium_account_update.sql
+++ b/transform/snowflake-dbt/models/hightouch/freemium/professional_subscription/freemium_account_update.sql
@@ -11,7 +11,7 @@ WITH customers_with_cloud_paid_subs as (
         'Customer' as account_type,
         -- Metadata to be used for selecting proper external id
         external_id_account.id as dwh_id_account_id,
-        external_id_account.id is not null as has_external_id_match,
+        external_id_account.id is not null as has_dwh_id_account_id,
         external_id_account.dwh_external_id__c as dwh_id_account_external_id,
 
         contact.accountid as contact_account_id,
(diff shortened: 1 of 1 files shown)


transform/snowflake-dbt/models/hightouch/freemium/professional_subscription/freemium_account_update.sql
```diff
@@ -8,14 +8,78 @@
 WITH customers_with_cloud_paid_subs as (
     SELECT
         customers_with_cloud_paid_subs.*,
-        'Customer' as account_type
+        'Customer' as account_type,
+        -- Metadata to be used for selecting proper external id
+        external_id_account.id as dwh_id_account_id,
+        external_id_account.id is not null as has_external_id_match,
```
catalintomai, line 14: why not `has_dwh_id_account_match` instead, to keep in sync with the previous renaming?

transform/snowflake-dbt/models/hightouch/freemium/professional_subscription/freemium_account_update.sql
```diff
@@ -8,14 +8,78 @@
...
+        external_id_account.dwh_external_id__c as dwh_id_account_external_id,
```
catalintomai, line 15: 0/5 - why keep the `dwh` lineage here and not below?

catalintomai (approved):

transform/snowflake-dbt/models/hightouch/freemium/professional_subscription/freemium_account_update.sql
ifoukarakis: Replaced.

PR #1281, opened 2023-05-25
Chore: more stripe sanity tests

ifoukarakis:
#### Summary

This PR focuses on onprem, self-serve subscriptions.

- [x] Add subscription items as they will be required for determining plan.
- [x] Add tests to validate items in https://mattermost.atlassian.net/wiki/spaces/DATAENG/pages/2453733377/Self+Serve+Renewals.
- [x] Validate license data are available in subscription metadata for onprem subscriptions.


transform/mattermost-analytics/tests/staging/stripe/test_subscription_items__max_two_per_onprem_subscription.sql
```diff
@@ -0,0 +1,15 @@
+-- Warn until the edge cases are clarified
+{{ config(
+    severity = 'warn',
+    error_if = '> 2'
+) }}
+
+select
```
catalintomai, line 7: Is this the place to address outlier cases like : https://mattermost.atlassian.net/wiki/spaces/DATAENG/pages/2453733377/Self+Serve+Renewals?focusedCommentId=2454552577 or should they be filtered out at the source in the staging model?

ifoukarakis: Hmmm, that's a good idea. Staging models are required in order to catch these issues, but we should avoid propagating them downstream. One thing that can be done is breaking staging model to `base` and `staging`. Any tests can rely on `base`, while `staging` can exclude known issues.

transform/mattermost-analytics/tests/staging/stripe/test_subscription__onprem_has_license_data.sql
```diff
@@ -0,0 +1,40 @@
+-- Test all specs defined in https://mattermost.atlassian.net/wiki/spaces/DATAENG/pages/2453733377/Self+Serve+Renewals
```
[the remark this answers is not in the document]
ifoukarakis: Agree. Renewal requires quite a few checks. However it requires having new subscriptions in place first. The plan is to merge this PR and start working on new subscriptions. As soon as this is done, renewals can be added.

transform/mattermost-analytics/tests/staging/stripe/test_subscription_items__max_two_per_onprem_subscription.sql
catalintomai: Trying to avoid having tests failures because of the "known issues".

ifoukarakis: This is actually a warning, not an error. It will be reported on DBT output, but will not break the build. If a 3rd case occurs any time in the future, the build will break. This will help us identify such new problematic cases.

Still, it's a good idea to have the check, then exclude the problematic cases from downstream models.

catalintomai (approved):

transform/mattermost-analytics/tests/staging/stripe/test_subscription_items__max_two_per_onprem_subscription.sql
catalintomai: Hear you, though I'd say even warnings are unnecessary noise for cases we know are invalid to begin with.

ifoukarakis: My idea was to use warnings as a type of monitoring. Let's discuss in our meeting possible options.

PR #1284, opened 2023-05-31
Removing license_server_fact from onprem_trial_request_facts

sri-byte:
Impact: Removing license_server_fact dependency from sources which use trial_requests.

Tables which will be updated here are 
- onprem_trial_request_facts

### DIFF
--- a/transform/snowflake-dbt/models/hightouch/product_campaigns/onprem_trial_request_facts.sql
+++ b/transform/snowflake-dbt/models/hightouch/product_campaigns/onprem_trial_request_facts.sql
@@ -20,10 +20,10 @@ with latest_server_daily_details as (
         tr.email
         , tr.site_url
         , tr.license_id
-        , license_server_fact.server_id
-        , license_server_fact.start_date
-        , license_server_fact.expire_date
-        , license_server_fact.edition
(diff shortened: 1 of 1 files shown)


transform/snowflake-dbt/models/hightouch/product_campaigns/onprem_trial_request_facts.sql
```diff
@@ -20,10 +20,10 @@ with latest_server_daily_details as (
...
         , tr.site_url
         , tr.license_id
-        , license_server_fact.server_id
-        , license_server_fact.start_date
-        , license_server_fact.expire_date
-        , license_server_fact.edition
+        , tr.server_id
+        , tr.start_date::date as start_date
+        , tr.end_date::date as expire_date
+        , 'E20 Trial' as edition
```
ifoukarakis, line 26: 3/5 this column is not needed downstream. Let's remove it.

sri-byte: dropped this column

sri-byte:
PR Closed, as the file has been deleted from the repo in another PR.

PR #1288, opened 2023-06-06
Creating model for contracted arr

RoyI99:

### DIFF
--- a/transform/snowflake-dbt/models/finance/contracted_arr_reporting.sql
+++ b/transform/snowflake-dbt/models/finance/contracted_arr_reporting.sql
@@ -1,0 +1,161 @@
+{{config({
+    "materialized": "table",
+    "schema": "finance",
+    "tags":'nightly'
+  })
+}}
+
+--v1 Roy CONTRACTED ARR REPORTING
+--tested on snowflake 20230606
+--this sql model reports ARR on date contracted instead of report date where report date is the later of close date or license start
+
+
+--population coming from arr_transactions
+with trans as (
+select 
+    t.* 
+    ,last_day(date_trunc('week',closing_date),'week') as closing_week
+    ,last_day(closing_date,month) as closing_month
+    ,last_day(dateadd('month',1,last_day(dateadd('month',2,date_trunc('quarter',dateadd('month',-1,closing_month)))))) as closing_fiscal_quarter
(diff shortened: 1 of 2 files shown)


transform/snowflake-dbt/models/finance/contracted_arr_reporting.sql
```diff
@@ -0,0 +1,161 @@
+{{config({
+    "materialized": "table",
+    "schema": "finance",
+    "tags":'nightly'
+  })
+}}
+
+--v1 Roy CONTRACTED ARR REPORTING
+--tested on snowflake 20230606
+--this sql model reports ARR on date contracted instead of report date where report date is the later of close date or license start
```
catalintomai, line 10: Is "date contracted" same as "date closed" (comments are slightly different in the two files).

catalintomai: Thanks, so for clarity, should the comments refer to/use the same terminology in the two files? Like "this sql model reports ARR on date contracted instead of report date" (this file) vs. "Same as arr_reporting but reporting date shifts to closing date instead of reporting date" (schema.yml)

RoyI99: Yes

RoyI99: @catalintomai it would be helpful if you directly ask the concern you are addressinf

catalintomai: @RoyI99 , it is more of a prevention of future/possible issues for me. If in the future someone new to MM and not familiar with this terminology will have to understand the logic here, they might ask why is the description in the schema file slightly different from the description of top of the file the schema file entry references (like the original q: is "date contracted" same as "closing date").

RoyI99: Yes zoom call is better

RoyI99: @catalintomai I pushed a second committ to clarify on the heading the difference between contract and close day.  Will document the rest on confluence.  Please approve so this will be part of the refresh tonight.   Thanks again!

catalintomai (approved):

PR #1292, opened 2023-06-14
MM-48421: active users model

ifoukarakis:
#### Summary

Creates a fact table that contains unique active users daily, weekly (rolling today + past 6 days) and monthly (rolling today and past 29 days).

- [x] Daily user activity
- [x] DAU
- [x] MAU
- [x] WAU

Note that this PR focuses on Rudderstack desktop/web data.

#### Ticket Link


sri-byte:
Hi @ifoukarakis, I'm validating the data with some `server_id`'s I have on hand.

transform/mattermost-analytics/models/intermediate/product/active_users/int_user_active_days_legacy_telemetry.sql
```diff
@@ -0,0 +1,45 @@
...
+    select
+        cast(received_at as date) as received_at_date,
+        cast(timestamp as date) as activity_date,
+        server_id,
+        user_id
+    from
+        {{ ref('stg_mattermost2__tracks') }}
+    where
+        -- Exclude items without user ids, such as server side telemetry etc
+        user_id is not null
```
catalintomai, line 22: Should we mention somewhere more prominently (like the folder name) that this work only includes webapp events (did a quick check for 06/16/2023 and out of 1.5M events about a third are eliminated by this check - prob same for non-legacy source as well)

transform/mattermost-analytics/models/staging/mattermost2/stg_mattermost2__tracks.sql
```diff
@@ -0,0 +1,18 @@
...
+        {{ ref ('base_mattermost2__tracks') }}
+)
+SELECT
+     id               AS event_id
+     , event          AS event_table
+     , event_text     AS event_name
+     , category       AS category
+     , type           AS event_type
+     , user_id        AS server_id
+     , user_actual_id AS user_id
```
catalintomai, line 15: couldn't find user_actual_id for raw.mattermost2.tracks though.

transform/mattermost-analytics/models/intermediate/product/active_users/int_user_active_days_legacy_telemetry.sql
```diff
@@ -0,0 +1,45 @@
...
+        -- Exclude items without server ids
+        and server_id is not null
```
catalintomai, line 24: Noticed that user_id being not NULL (renamed here as server_id) does not guarantee that these are analytics.js events only, there are cases where both analytics.js and analytics-go libraries generate such events.

transform/mattermost-analytics/models/staging/mattermost2/stg_mattermost2__tracks.sql
sri-byte: The `user_actual_id` for mattermost2 is in `mattermost2.event` table

sri-byte:
LGTM so far

transform/mattermost-analytics/models/staging/mattermost2/stg_mattermost2__tracks.sql
catalintomai: Yes, my bad, `get_event_relations()` gets them from all `rudderstack_tables`.

transform/mattermost-analytics/models/intermediate/product/active_users/int_user_active_days_legacy_telemetry.sql
ifoukarakis: As far as I understand, `server_id` (coming from `user_id`) must not be `null`, as it's expected to be there to denote the server that the event is originating from. If it is equal to null, then it's some special case that needs to be investigate.

Regarding `user_id` (coming from `user_actual_id`), the goal was to keep only user generated events, and not server generated. I agree that this can be further optimized to exclude such events earlier in the pipeline. However, I think it's a good point for the second iteration. The reasons are:
- Current full refresh process took 1 hour. Daily refreshes should take way less.
- The optimization would require excluding specific event tables that do not have a column named `user_actual_id`. This is a bit tricky to implement. A possible proxy is to implement a macro that iterates over the tables in the schema and gets the ones with both a `user_id` and `user_actual_id` column.  This can be a bit challenging to implement, and might grow the size of the current PR a bit.

transform/mattermost-analytics/models/staging/mattermost2/stg_mattermost2__tracks.sql
ifoukarakis: @catalintomai exactly, it's the same idea as with rudderstack data. `user_actual_id` is a custom column, thus never added in `tracks` table. The workaround is to create a "synthetic" tracks view that unions all event tables. Note that the synthetic tracks table is a view, and not a table. The view can be always updated to include new tables, without a need to do a full refresh.

transform/mattermost-analytics/models/intermediate/product/active_users/int_user_active_days_legacy_telemetry.sql
ifoukarakis: Also note that these synthetic tables can help answering questions like daily count of events per server/user etc. This seems to be a recurring pattern from product.

sri-byte (approved):

transform/mattermost-analytics/models/intermediate/product/active_users/int_user_active_days_legacy_telemetry.sql
catalintomai: "the goal was to keep only user generated events, and not server generated" - might be a good q to ask ourselves - what semantics are we missing (if any) by excluding the system-generated ones - like is the line between the two sets blurry or not? Could we potentially (now or in a future PR) tag them as such (like an `is_system` flag)  to be able to filter them out/build metrics targeting them specifically.

sri-byte (approved):

PR #1299, opened 2023-06-23
Hotfix: Ignore record with invalid data

ifoukarakis:
#### Summary

These records seems to be from a security researcher. Values are strange. For example first name is `contact_first_name` and last name is `contact_last_name`, while company size is equal to `100,000,000`.

Manually ignoring those records for now.


sri-byte (approved):

catalintomai:
@ifoukarakis , do you think we should cc security, they might keep want to be in the loop. Thanks.

ifoukarakis:
> 

I'll drop a line in the related channel and move forward depending on their suggestions.

PR #1305, opened 2023-06-27
Hotfix: fix bearer prefix for hightouch authentication

ifoukarakis:
#### Summary

Hightouch uses [bearer auth](https://hightouch.com/docs/api-reference#section/Authentication/bearerAuth). Current setup was pushing `Bearer ` prefix as part of the secret. Updating the code to make the need for this explicit and simplify secret management.

### DIFF
--- a/dags/general/monitoring.py
+++ b/dags/general/monitoring.py
@@ -45,7 +45,7 @@ with DAG(
         http_conn_id="hightouch",
         method="GET",
         endpoint=Variable.get('hightouch_syncs_endpoint'),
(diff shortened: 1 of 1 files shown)


sri-byte (approved):

catalintomai:
@ifoukarakis , does that mean that the `hightouch_secret` needs to be updated/trimmed to not include the `Bearer` prefix that has been moved out?

ifoukarakis:
> @ifoukarakis , does that mean that the `hightouch_secret` needs to be updated/trimmed to not include the `Bearer` prefix that has been moved out?

Exactly. This is already done in the new terraform configuration and tested on the new resources.

PR #1308, opened 2023-06-28
MM-48421: adding daily server info snapshot

ifoukarakis:
#### Summary

- [x] Add daily server info snapshot
- [x] Add excluded servers
- [x] Create fact and dimension tables

Note that the current excluded servers implementation is a first approach. More elaborate solutions can be added. 

#### Ticket Link


transform/mattermost-analytics/models/intermediate/product/servers/_latest_daily/int_server_telemetry_latest_daily.sql
```diff
@@ -21,7 +16,8 @@ select
     database_version,
     edition,
     -- Can be used to identify potential upgrade/upgrade attempts or erroneous data
```
sri-byte, line 18: can we please also add `max(version_full)` as well

ifoukarakis: Can you please elaborate a bit more? You want the maximum version for a specific date or for the whole history of the server?

Also since it's string comparison, it's a bit tricky to use `max`.

sri-byte: I'd like to see what a server's latest version is, at the current time. If `max` won't work here, because its a string, can we do `last(distinct version_major) over (partition by server_id, server_date desc) as last_server_version`. This will show us the latest version of the server. (Also, corrected to version_major, from version_full)

ifoukarakis: This table holds exactly that: the latest version reported in the day. This line of the query makes sure about it: https://github.com/mattermost/mattermost-data-warehouse/blob/master/transform/mattermost-analytics/models/intermediate/product/servers/int_server_telemetry_latest_daily.sql#L28 (it was there from the previous PR).

Here's some sample scenarios:
- Server reported only one version (i.e. 5.2.3) for `2022-01-01`. Then `version_full` will be `5.2.3` and version major/minor/patch will be 5, 2, 3 respectively.
- Server was upgraded during the day, reporting `5.2.3` one time, and then new version `6.0.1`. Note that on server restart, a version might be reported multiple times. The latest version reported for the day will be `6.0.1`.
- Server was on version `5.2.3` at the start of the day, then upgraded to `6.0.1`. Upgrade failed and server was rolled back to `5.2.3`. All operations happened on the same day. Latest version reported  for the day will be `5.2.3`.
- Same scenario as the last one, but after rollback to `5.2.3` it was upgraded to `5.4.2`. `version_full` will be `5.4.2`.

If something more is needed, let's do it within the scope of the related JIRA ticket. We can add other things there, based on what we need. In the meantime, let's proceed with this PR, unless there are other issues.

catalintomai (approved):

PR #1324, opened 2023-07-26
Hotfix: Add missing country code to seed file

sri-byte:
Impact: dbt hourly is failing due to a missing country code in the seed file.


catalintomai:
@sri, let's just change the description+title to make it clear that the country code is correct (the name is incorrect). Also, recommend including the link to the the bug on Nick/other team. Thanks.

sri-byte:
Updated description to reflect the issue.

catalintomai (approved):

PR #1329, opened 2023-08-09
MM-48421: server activity metrics

ifoukarakis:
#### Summary

Enrich active users fact with active user counts from server-side data telemetry.

#### Ticket Link

### DIFF
--- a/transform/mattermost-analytics/models/intermediate/product/servers/_latest_daily/int_activity_latest_daily.sql
+++ b/transform/mattermost-analytics/models/intermediate/product/servers/_latest_daily/int_activity_latest_daily.sql
@@ -1,7 +1,7 @@
 select
     server_id,
     CAST(timestamp AS date) AS server_date,
-    {{ dbt_utils.generate_surrogate_key(['server_id', 'server_date']) }} AS daily_server_id,
+    {{ dbt_utils.generate_surrogate_key(['server_date', 'server_id']) }} AS daily_server_id,
     daily_active_users,
     monthly_active_user,
     count_registered_users,
--- a/transform/mattermost-analytics/models/intermediate/product/servers/_latest_daily/int_activity_latest_daily.sql
+++ b/transform/mattermost-analytics/models/intermediate/product/servers/_latest_daily/int_activity_latest_daily.sql
@@ -1,7 +1,7 @@
 select
     server_id,
     CAST(timestamp AS date) AS server_date,
-    {{ dbt_utils.generate_surrogate_key(['server_date', 'server_id']) }} AS daily_server_id,
+    {{ dbt_utils.generate_surrogate_key(['server_id', 'server_date']) }} AS daily_server_id,
     daily_active_users,
     monthly_active_user,
     count_registered_users,
--- a/transform/mattermost-analytics/models/intermediate/product/servers/_latest_daily/int_activity_latest_daily.sql
+++ b/transform/mattermost-analytics/models/intermediate/product/servers/_latest_daily/int_activity_latest_daily.sql
@@ -5,7 +5,7 @@ select
     daily_active_users,
     monthly_active_user,
     count_registered_users,
-    registered_deactivated_users
+    count_registered_deactivated_users
 from
     {{ ref('stg_mm_telemetry_prod__activity') }}
 where
--- a/transform/mattermost-analytics/models/intermediate/product/servers/_latest_daily/int_activity_latest_daily.sql
+++ b/transform/mattermost-analytics/models/intermediate/product/servers/_latest_daily/int_activity_latest_daily.sql
@@ -3,7 +3,7 @@ select
     CAST(timestamp AS date) AS server_date,
     {{ dbt_utils.generate_surrogate_key(['server_id', 'server_date']) }} AS daily_server_id,
     daily_active_users,
-    monthly_active_user,
+    monthly_active_users,
     count_registered_users,
     count_registered_deactivated_users
 from
--- a/transform/mattermost-analytics/models/intermediate/product/servers/_latest_daily/int_activity_latest_daily.sql
+++ b/transform/mattermost-analytics/models/intermediate/product/servers/_latest_daily/int_activity_latest_daily.sql
@@ -1,0 +1,15 @@
+select
+    server_id,
+    CAST(timestamp AS date) AS server_date,
+    {{ dbt_utils.generate_surrogate_key(['server_id', 'server_date']) }} AS daily_server_id,
+    daily_active_users,
+    monthly_active_user,
+    count_registered_users,
+    registered_deactivated_users
+from
+    {{ ref('stg_mm_telemetry_prod__activity') }}
+where
+    -- Ignore rows where server date is in the future.
+    server_date <= CURRENT_DATE()
+-- Keep latest record per day
+qualify row_number() over (partition by server_id, server_date order by timestamp desc) = 1
--- a/transform/mattermost-analytics/models/marts/product/fct_active_users.sql
+++ b/transform/mattermost-analytics/models/marts/product/fct_active_users.sql
@@ -25,7 +25,7 @@ select
     m.*,
     -- Server activity
     coalesce(sas.daily_active_users, 0) as server_daily_active_users,
-    coalesce(sas.monthly_active_user, 0) as server_monthly_active_users,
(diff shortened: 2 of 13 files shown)


transform/mattermost-analytics/models/intermediate/product/servers/_latest_daily/int_activity_latest_daily.sql
```diff
@@ -0,0 +1,15 @@
+select
+    server_id,
+    CAST(timestamp AS date) AS server_date,
+    {{ dbt_utils.generate_surrogate_key(['server_id', 'server_date']) }} AS daily_server_id,
+    daily_active_users,
+    monthly_active_user,
```
catalintomai, line 6: `monthly_active_user**s**`?

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__models.yml
```diff
@@ -116,3 +116,74 @@ models:
...
+      - name: event_id
+        description: The event's id.
+      - name: event_name
+        description: The name of the event.
+      - name: event_table
+        description: The name of the event table.
+      - name: server_id
+        description: The ID of the server the event originated from.
+      - name: received_at
+        description: Timestamp registered by Segment when the event was ingested (received).
```
catalintomai, line 133: replace Segment with Rudderstack? (here and below)?

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__activity.sql
```diff
@@ -0,0 +1,73 @@
...
+        , posts_previous_day as count_posts_previous_day
+        , teams as count_teams
+        , slash_commands as count_slash_commands
+        , direct_message_channels as count_direct_message_channels
+        , posts as count_posts
+        , incoming_webhooks as count_incoming_webhooks
+        , outgoing_webhooks as count_outgoing_webhooks
+
+        -- User & account metrics
+        , active_users_daily as daily_active_users
```
catalintomai, line 37: maybe to open a bug to investigate at a certain moment why `active_users_daily` is populated and `daily_active_users` not

ifoukarakis: There were only ~ 800 records (out of 100m+) that had a value in `daily_active_users`. The [source code of mattermost server](https://github.com/mattermost/mattermost/blob/master/server/platform/services/telemetry/telemetry.go#L379) is using `active_users_daily`. I even checked versions back to 4.x and the code seems to be using this alias since then. My assumptions is that it's either a test/dev server, or part of some pen test. Not sure if it's worth spending time to investigate this.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__activity.sql
```diff
@@ -0,0 +1,73 @@
...
+        , active_users_monthly as monthly_active_user
+        , registered_users as count_registered_users
+        , registered_deactivated_users as count_registered_deactivated_users
+        , bot_accounts as bot_accounts
+        , guest_accounts as guest_accounts
+
+        -- Metadata from Rudderstack
+        , context_library_version
+        , context_library_name
+        , uuid_ts
```
catalintomai, line 47: maybe update the naming conventions in https://mattermost.atlassian.net/wiki/spaces/DATAENG/pages/2422603777/DWH+Naming+Conventions that `*_ts` is acceptable as well for timestamps - like `timestamp` (or rename this one)

ifoukarakis: I was between keeping the value as is or completely dropping it. It's never used downstream. Also according to [Rudderstack's docs](https://www.rudderstack.com/docs/destinations/warehouse-destinations/warehouse-schema/) "Added by RudderStack for debugging purposes. Can be ignored for analytics."

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__activity.sql
catalintomai: That's fine - then maybe we should just add `daily_active_users` to a "to remove/not ingest" list to clean-up at a certain point- to avoid extreme sparseness/empty columns.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__activity.sql
catalintomai: I think so, we can always add it necessary.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__activity.sql
ifoukarakis: I'm kind of documenting these things by leaving comments on the staging models about my investigation on these columns (here's an [example](https://github.com/mattermost/mattermost-data-warehouse/pull/1329/files#diff-ad95894abfb0112a643b913535b5b52076d7b7a0970400d1f58498d35f034f52R53-R55)). Do you think we need something more?

transform/mattermost-analytics/models/marts/product/fct_active_users.sql
```diff
@@ -21,10 +21,18 @@ with metrics as (
     group by activity_date, server_id
 )
 select
-    *
+    -- Client telemetry
+    m.*,
+    -- Server activity
+    coalesce(sas.daily_active_users, 0) as server_daily_active_users,
+    coalesce(sas.monthly_active_user, 0) as server_monthly_active_users,
```
catalintomai, line 28: `monthly_active_user` -> `monthly_active_users` here as well since you made the change in the other places.

ifoukarakis: Nice catch. CI/CD failed because of that.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__models.yml
```diff
@@ -116,3 +116,74 @@ models:
...
+        description: Total number of incoming webhooks, excluding deleted.
+      - name: count_outgoing_webhooks
+        description: Total number of outgoing webhooks, excluding deleted.
+      - name: daily_active_users
+        description: |
+          Total number of active users for the past 24 hours prior to the timestamp of the event.
+          Bots and deleted users are excluded.
+      - name: monthly_active_users
+        description: |
+          Total number of active users for the past 31 days prior to the timestamp of the event.
```
catalintomai, line 173: the `metrics` CTE in `fct_active_users.sql` is using a 30 days window - 
sum({{metric}}_last_30_days::integer) as monthly_{{column}} - not sure if we need to be aligned on 30 or 31 days across.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__activity.sql
catalintomai: Not really, as long as we can easily do a scrub to find them when the time comes = ).

transform/mattermost-analytics/models/staging/mm_telemetry_prod/_mm_telemetry_prod__models.yml
ifoukarakis: Not sure as well. 31 is the max number but 30 is kind of a balancing item. 

Another idea is that if they metrics need to be super accurate, then the window should use activity event timestamp as reference. But this can become challenging in implementation.

catalintomai: I think that's fine then. Maybe just something to keep in mind if we detect small inaccuracies later on.

catalintomai (approved):

PR #1333, opened 2023-08-14
MM-48421: fix bug with null being return in select in

ifoukarakis:
#### Summary

Intermediate table might contain `null` as server id, causing `WHERE id NOT IN (SELECT server_id FROM...)` to exclude all rows.

### DIFF
--- a/transform/mattermost-analytics/models/intermediate/product/servers/int_excludable_servers.sql
+++ b/transform/mattermost-analytics/models/intermediate/product/servers/int_excludable_servers.sql
@@ -15,8 +15,8 @@
 )
 select * from seed_file
 union all
-select * from {{ ref('int_excludable_servers_invalid_security_data') }}
+select * from {{ ref('int_excludable_servers_invalid_security_data') }} where server_id is not null
 union all
-select * from {{ ref('int_excludable_servers_cloud_installations') }}
+select * from {{ ref('int_excludable_servers_cloud_installations') }} where server_id is not null
 union all
-select * from {{ ref('int_excludable_servers_single_day_activity') }}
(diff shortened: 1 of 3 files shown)


sri-byte (approved):

transform/mattermost-analytics/models/intermediate/product/servers/int_excludable_servers.sql
```diff
@@ -15,8 +15,8 @@
 )
 select * from seed_file
 union all
-select * from {{ ref('int_excludable_servers_invalid_security_data') }}
+select * from {{ ref('int_excludable_servers_invalid_security_data') }} where server_id is not null
```
catalintomai, line 18: Yiannis, as a general pattern/recommendation, should this null checks be here or at the sources (e.g. `int_excludable_servers_invalid_security_data`) ? Might not matter here too much, but for the future. (reason I'm asking is because I believe it is generally better to aggregate the filtering logic in one place as opposed to spread it across the files, but don't know if there are any "official" recommendations).

ifoukarakis: I don't have a strong opinion. My thoughts are:
- Most of the tables involved are in the intermediate layer, thus they might need to offer both versions of the data.  
- Basic cleaning should happen as close to the source as possible (staging layer). Having said that, rows with null values might have a special meaning and need to be excluded only for some use cases.
- Nulls are a special case in SQL, I think it wouldn't hurt to be "defensive" and add the extra check whenever needed. 
- If we need to have a contract that a column doesn't contain nulls, we should also add a test for that in order to enforce it. For the specific use case looks like too much of an effort and additional queries with minimum gain.

Let's see if we have other cases like that and we can decide if we can derive a rule.

PR #1336, opened 2023-08-21
MM-51877 Create github action seed file

sri-byte:
Impact: There is a CSV file that can be exported from Github Actions. The team would like this data to be available on Looker in order to create some reports. This is their of our Observability on CI/CD goal. 

This PR, creates a seed file under seed/github folder, to which Antonis and team will append the data manually every 15 days.

#### Ticket Link
  https://mattermost.atlassian.net/browse/MM-51877


toninis:
Do we support also deduplication ?

transform/mattermost-analytics/models/intermediate/github_actions/int_github_actions.sql
```diff
@@ -0,0 +1,25 @@
...
+        , trim(multiplier) as multiplier
+        , trim(owner) as owner
+        , trim(repository_slug) as repository_slug
+        , trim(username) as username
+        , trim(actions_workflow) as actions_workflow
+        , trim(notes) as notes
+    from
+        {{ ref('github_actions') }}
+)
+select distinct * from seed_file
```
sri-byte, line 25: @toninis deduplication is handled here by using `distinct`

PR #1337, opened 2023-08-23
MM-54215: split merged dimensions

ifoukarakis:
#### Summary

- [x] Split version dimension from daily server info. Version was split as a separate dimension while working on the active server models, thus it makes sense to avoid duplication.
- [x] More elaborate logic for checking the hosting type of a server.  Since this is something that's hard to change, it makes sense to move it to a type-1 dimension. The new logic compares the number of days that a server reports an installation id (present only on cloud hosted servers) vs the ones that it reports it as null. 

⚠️  Note that merging this PR will break existing views/explores.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54215


transform/mattermost-analytics/models/marts/product/_product__models.yml
```diff
@@ -240,3 +221,19 @@ models:
...
+        tests:
+          - not_null
+      - name: installation_id
+        description: The id of the installation. Exists only for cloud installations.
+      - name: hosting_type
+        description: |
+          Type of hosting for this server. Can be one of:
+          - `Cloud`: server always reports an installation id
+          - `Self-hosted`: server always reports a null installation id.
+          - `Probably cloud`: server reports an installation id more often than reporting null as installation id.
```
catalintomai, line 238: for 'Probably Cloud' you compare against `count_not_is_cloud_days` not against `count_unknown_is_cloud_days` (which covers the nulls). Same for `Probably self-hosted`.

transform/mattermost-analytics/models/marts/product/dim_server_info.sql
```diff
@@ -0,0 +1,35 @@
...
+        server_id
+), latest_values as (
+    -- Get latest values for installation id (if exists) and full version string
+    select
+        server_id,
+        installation_id
+    from
+         {{ ref('int_server_active_days_spined') }}
+    -- Keep latest record per day
+    qualify row_number() over (partition by server_id order by activity_date desc) = 1
```
sri-byte, line 21: instead of latest record, if we use the last value where it was `not null` we can get the `installation_id`, since this is a dimension table.

transform/mattermost-analytics/models/marts/product/_product__models.yml
ifoukarakis: It's a bit tricky. In the date spine per server id table, a row exists for each combination of server id and every date between the first and last date any telemetry/security update exists. Unfortunately there is no guarantee that data exists for each date. For example, a server might have sent it's first telemetry on `2023-08-01` and it's last on `2023-08-07`, but telemetry is missing on `2023-08-03`. It's those cases where `is_cloud` is null, as there is no data and it's impossible to check if the server has an installation id or not. 

The check for deciding if the server is cloud or not uses the following logic:
- Count the number of days where the server is reporting it's a cloud server.
- Count the number of days where the server is reporting it's a self-hosted server.
- If the server reports only cloud server days, then it's definitely a cloud server.
- If the server reports only self-hosted server days, then it's definitely a self-hosted server.
- If the server reports both self-hosted and cloud, use the largest value of the two, but mark it as `Probably`.
- If there's no information regarding installation id in activity data, it's impossible to decide if it's a cloud or on prem server based on installation id. A more elaborate solution should be added.

A few final remarks:
- `count_unknown_is_cloud_days` is not used. Perhaps it should be removed for avoiding confusion.
- The majority of the servers reports either cloud or self hosted days. There are only ~ 43 servers that report both, and should be investigated as special cases.

transform/mattermost-analytics/models/marts/product/dim_server_info.sql
ifoukarakis: The idea of this CTE is to keep the latest installation id in case the server reports multiple installation ids. There are a few cases that this happens.

ifoukarakis:
> 

Stripe data contain installation id but no server id. It's going to be a significant effort to connect the two (most likely it will be part of the next thing we'll be working on).

ifoukarakis:
> 


Stripe data contain installation id but no server id. It's going to be a significant effort to connect the two (most likely it will be part of the next thing we'll be working on).

ifoukarakis:
> For cloud there is no option for the customer/server to turn off telemetry, if the data is missing it is likely some code/timing/unknown issue but not an issue of reporting telemetry.

This is unfortunately happening. That's one of the reasons this PR attempts to use the full history of the server activity in order to decide whether the server is hosted in cloud or on-prem.

transform/mattermost-analytics/models/marts/product/_product__models.yml
catalintomai: Thanks for the detailed answer. Made the comment because you mentioned in the description: "server reports an installation id more often than reporting null" (and I thought that `count_unknown_is_cloud_days ` captures this semantics) as opposed to "server reports both self-hosted and cloud, use the largest value of the two". Agree that better not have `count_unknown_is_cloud_days` if not used (or replace it with a TODO statement instead)

ifoukarakis: Removed.

transform/mattermost-analytics/models/marts/product/_product__models.yml
```diff
@@ -240,3 +217,19 @@ models:
...
+        tests:
+          - not_null
+      - name: installation_id
+        description: The id of the installation. Exists only for cloud installations.
+      - name: hosting_type
+        description: |
+          Type of hosting for this server. Can be one of:
+          - `Cloud`: server always reports an installation id
+          - `Self-hosted`: server always reports a null installation id.
+          - `Probably cloud`: server reports an installation id more often than not reporting as installation id.
```
sri-byte, line 234: IMO, we shouldn't have these 2 text (`Probably cloud` and `Probably self-hosted` in the data, its speculative and might leave the end user confused, until we surely know if its `Cloud` or `Self-Hosted`, which we are going to in the near future, combining other sources of data, we should leave this blank.

sri-byte (approved):

transform/mattermost-analytics/models/marts/product/_product__models.yml
ifoukarakis: Thank you for the feedback. Should `Probably cloud`, `Probably self-hosted` and null values be replaced with something like `Unknown`? We can also use the number of rows with that value as a quality metric.

PR #1360, opened 2023-10-20
MM-54964: include excludable servers

ifoukarakis:
#### Summary

Handling inconsistency and bugs found while comparing current solution with company score card.
 
- [x] Don't ignore excludable servers in active users fact. This can be done in Looker in order to provide finegrained analysis.
- [x] BUGFIX: Server-reported DAU/MAU wasn't appearing in the table if there was no user telemetry.

> After this PR has been merged, Looker explore for active users will need to be updated to include excludable servers.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54964


transform/mattermost-analytics/models/marts/product/fct_active_users.sql
```diff
@@ -22,20 +22,28 @@ with metrics as (
 )
 select
     -- Client telemetry
-    m.*,
+    coalesce(m.daily_server_id, sas.daily_server_id) as daily_server_id
+    , coalesce(m.activity_date, sas.activity_date) as activity_date
+    , coalesce(m.server_id, sas.server_id) as server_id
+    {% for metric, column in column_map.items() %}
```
catalintomai, line 28: q: do we need the for loop here as well since we had one in the `metrics` CTE above?

catalintomai (approved):

PR #1361, opened 2023-10-23
Hotfix: Add country to exclusion list as its failing hourly job.

sri-byte:
Impact: Add country to exclusion list as its failing hourly job.


ifoukarakis:
Country names will be replaced by country code. I'll open a separate ticket to work on it.

PR #1368, opened 2023-10-30
Add actual release dates

amyblais:
The file currently lists planned release dates, I'm also adding actual release dates after discussing with Yiannis.

I've added the actual release dates for the past year, but let me know if I should fill in the details for all versions.


ifoukarakis (approved):
Approving. It would be nice if @catalintomai or @sri-byte could have a look!

catalintomai (approved):

PR #1370, opened 2023-10-31
MM-55026: refactor releaes model

ifoukarakis:
#### Summary

Refactor release model. 
- [x] Split release from fix version dimension. The main reason is that a lot of reports rely on the release (and its metadata) rather than the fix version. Also release has a different granularity than fix version.
- [x] Tag fix version based on their text pattern in order to identify if it's a self-hosted (AKA on-prem) fix version or cloud version.
- [x] Release dimension contains extra metadata.
- [x] Minor adjustments to match recommended style. 
- [x] Extra tests for conventions on values.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-55026

### DIFF
--- a/transform/mattermost-analytics/models/marts/release/dim_fix_versions.sql
+++ b/transform/mattermost-analytics/models/marts/release/dim_fix_versions.sql
@@ -3,12 +3,12 @@
 -- Implement filtering on this layer as it's only used here.
 select
     issue_id,
-    value:name::string as fix_version_name,
+    value:name::string as fix_version,
     -- Break down different variations of target version
-    regexp_substr(fix_version_name, 'v\\d+\.\\d+') as semver,
-    regexp_substr(fix_version_name, 'v(\\d+)', 1, 1, 'e', 1)::int as version_major,
-    regexp_substr(fix_version_name, '\\.(\\d+)', 1, 1, 'e', 1)::int as version_minor,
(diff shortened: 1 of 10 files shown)


transform/mattermost-analytics/models/marts/release/dim_fix_versions.sql
```diff
@@ -1,42 +1,32 @@
...
+    case
+        when fix_version ilike '%mobile%' then 'Mobile'
+        when fix_version ilike '%desktop%' then 'Desktop'
+        when fix_version ilike '%playbooks%' then 'Playbooks'
+        when fix_version ilike '%ir%' then 'IR'
+        when fix_version ilike '%cloud%' then 'Cloud'
+        when fix_version ilike '%apps%' then 'Apps'
+    end as component,
+    to_date(regexp_substr(fix_version, '\\d{2}/\\d{2}/\\d{2}'), 'mm/dd/yy') as cloud_release_date,
+    fix_version like any (
```
catalintomai, line 21: would be nice to document/give an example to how a self-hosted `fix_version` would look like vs a cloud one.

catalintomai (approved):

transform/mattermost-analytics/models/marts/release/dim_fix_versions.sql
ifoukarakis: Added in docs.

PR #1376, opened 2023-11-03
MM-55026: move next release to release

ifoukarakis:
#### Summary

Moving the logic for next/previous release to `dim_release` as it seems it's the part of the query that's needed.
Also handling patch versions for detecting next short version (major and minor), ignoring patch releases.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-55026

### DIFF
--- a/transform/mattermost-analytics/models/marts/release/dim_releases.sql
+++ b/transform/mattermost-analytics/models/marts/release/dim_releases.sql
@@ -1,4 +1,17 @@
+with release_versions as (
+    select
+        short_version,
+        lag(short_version, 1) over(order by release_number asc) as previous_release_version,
+        lag(short_version, -1) over(order by release_number asc) as next_release_version
+    from
+        {{ ref('stg_mattermost__version_release_dates') }}
+    where
+        rc1_date is not null
+)
 select
-    {{ dbt_utils.star(ref('stg_mattermost__version_release_dates'))}}
+    {{ dbt_utils.star(ref('stg_mattermost__version_release_dates'))}},
+    rv.previous_release_version,
+    rv.next_release_version
 from
-    {{ ref('stg_mattermost__version_release_dates') }}
+    {{ ref('stg_mattermost__version_release_dates') }} rd
+    join release_versions rv on rd.short_version = rv.short_verion
--- a/transform/mattermost-analytics/models/marts/release/_release__models.yml
+++ b/transform/mattermost-analytics/models/marts/release/_release__models.yml
@@ -95,16 +95,7 @@ models:
         description: |
           Whether this fix version refers to a cloud release. Examples of Cloud fix versions are 
           `Cloud 02/09/23 -> v7.9` and `Cloud 4/27/22 -> v6.7 May'22`.
-      - name: previous_release_version
-        description: |
-          The version of the last release before the fix version's release. Patch versions are ignored. 
-          For example, let's assume that known versions are `7.8`, `7.9`, `7.10`, `8.0` and `8.1`. Previous version of
-          `7.9` is `7.8`, previous version of `8.0` is `7.10` etc.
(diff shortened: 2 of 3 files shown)


transform/mattermost-analytics/models/marts/release/dim_releases.sql
```diff
@@ -1,4 +1,17 @@
+with release_versions as (
+    select
+        short_version,
+        lag(short_version, 1) over(order by release_number asc) as previous_release_version,
+        lag(short_version, -1) over(order by release_number asc) as next_release_version
+    from
+        {{ ref('stg_mattermost__version_release_dates') }}
+    where
```
catalintomai, line 8: I assume `version_patch = 0` filter was superfluous in the prev. iteration?

ifoukarakis: It's pretty much the same with releases having an RC1 date. 

Perhaps we can also add a test @ the source checking that whenever rc1 date is not null, patch_version is equal to 0.

transform/mattermost-analytics/models/marts/release/fct_issues_daily_snapshot.sql
```diff
@@ -1,14 +1,44 @@
...
+    -- Keep releases
+    select
+        version,
+        short_version,
+        rc1_date,
+        planned_release_date,
+        actual_release_date,
+        -- Calculate 17th of month before the release's month. This is required as release timeframe is 17th to planned
+        -- release date.
+        dateadd(month, -1, planned_release_date) as _month_before_release,
```
catalintomai, line 11: this prob. need documenting. For `_month_before_release` we use `planned_release_date` and for `week_after_release` we are using the `actual_release_date`

ifoukarakis: It is already documented in https://mattermost.atlassian.net/wiki/spaces/DATAENG/pages/2490171393/Release+Quality. It is also documented in the additions in `_release_models.yml`. Do you think something extra is needed?

ifoukarakis: Documentation was an extra reason for moving this logic from looker to DBT. Things seem easier to be documented here.

catalintomai: since we do the `coalesce(actual_release_date::date, planned_release_date) as actual_release_date` in the staging layer, maybe it makes sense to use `actual_release_date` exclusively?

ifoukarakis: No, it's two different things. 

- Planned release date is the date that the release is originally planned for. It is always present in the CSV.
- Actual release date is the date the release happened. It might >=  than the planned release date.  If it's `=`, then it's considered a normal release. If it's `>`, then it's considered a late release.
- CSV file contains some of the entries for actual dates. Some of them are equal to the planned release date, some are not and some are empty. For the ones that are empty, we consider the actual release date as equal to planned release date. 
- There's three important timeframes.
  - 17th to planned release date.
  - RC1 cut date to planned release date (the last part of the previous one).
  - A week after actual release date. Note that depending on whether a release is late or not, there might be a "gap" between planned release date and this timeframe.

The diagram in the [confluence](https://mattermost.atlassian.net/wiki/spaces/DATAENG/pages/2490171393/Release+Quality#Language) might also help understand this.

catalintomai: ok, I think I get it - it is by design. I think documentation is great, should've read it before.

catalintomai (approved):

PR #1377, opened 2023-11-03
Chore: Creates new intermediate model int_server_hosting_type 

sri-byte:
Impact: Creates a new intermediate model `int_server_hosting_type` so the logic for hosting type can be reused downstream in other dim and fact models.

### DIFF
--- a/transform/mattermost-analytics/models/intermediate/product/servers/int_server_hosting_type.sql
+++ b/transform/mattermost-analytics/models/intermediate/product/servers/int_server_hosting_type.sql
@@ -9,14 +9,23 @@ with server_telemetry_summary as (
         {{ ref('int_server_active_days_spined') }}
     group by
         server_id
+), user_telemetry_summary as (
+    select
+        server_id
+    from
+        {{ ref('int_user_active_days_spined') }}
+    where
+        is_active_today
+    group by 
+        server_id
 ), server_info as (
    select
        coalesce(st.server_id, ut.server_id) as server_id,
        coalesce(st.count_is_cloud_days, 0) as count_is_cloud_days,
        coalesce(st.count_not_is_cloud_days, 0) as count_not_is_cloud_days
     from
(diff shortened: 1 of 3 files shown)


ifoukarakis (changes requested):
`dim_server_info` on the branch contains info where `hosting_type` is null. This is not happening on the master branch, so a bug must have been introduced.

Also checked if there are differences in the hosting type between old and new approach. Only 3 servers seem to have a different hosting type.

transform/mattermost-analytics/models/intermediate/product/servers/int_server_hosting_type.sql
```diff
@@ -11,11 +11,12 @@ with server_telemetry_summary as (
         server_id
 ), server_info as (
    select
-       server_id as server_id,
+       coalesce(st.server_id, ut.server_id) as server_id,
        coalesce(st.count_is_cloud_days, 0) as count_is_cloud_days,
        coalesce(st.count_not_is_cloud_days, 0) as count_not_is_cloud_days
     from
-        server_telemetry_summary st
+        server_telemetry_summary st full outer join
```
catalintomai, line 18: nit: move "full outer join" statement on next row (in front of the "ref"), to be consistent across the codebase.

sri-byte: Updated.

sri-byte:
> `dim_server_info` on the branch contains info where `hosting_type` is null. This is not happening on the master branch, so a bug must have been introduced.
> 
> Also checked if there are differences in the hosting type between old and new approach. Only 3 servers seem to have a different hosting type.

This is now updated, I missed a join to the user telemetry data which is now added.

sri-byte:
The results now align with master

catalintomai (approved):

ifoukarakis (approved):

PR #1379, opened 2023-11-07
MM-54964: remove early condition for excludable servers

ifoukarakis:
### Summary

- [x] Remove exclusion of excludable servers from `dim_daily_server_info`, as it should be performed by joining with `dim_excludable_servers`.
- [x] Label server `E0` and `TE`.
- [x] Document `E0` and `TE`.

Note that `E0`/`TE` labels are applied in the intermediate label as we might need this information to calculate the first and current server's binary edition.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54964

### DIFF
--- a/transform/mattermost-analytics/models/intermediate/product/servers/_int_servers__models.yml
+++ b/transform/mattermost-analytics/models/intermediate/product/servers/_int_servers__models.yml
@@ -34,6 +34,11 @@ models:
         description: The version of the database.
       - name: is_enterprise_ready
         description: Whether this server is an enterprise-ready build.
+      - name: binary_edition
+        description: |
+          There are two "flavours" of the server binary:
+          - TE = Team Edition (fully open source and contains no Mattermost proprietary code).
+          - E0 = Enterprise Edition (TE + Mattermost proprietary code, and with features activated with a license).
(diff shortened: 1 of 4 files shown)


transform/mattermost-analytics/models/intermediate/product/servers/_int_servers__models.yml
```diff
@@ -34,6 +34,11 @@ models:
         description: The version of the database.
       - name: is_enterprise_ready
         description: Whether this server is an enterprise-ready build.
+      - name: binary_edition
```
catalintomai, line 37: Do you think this will evolve into a more-encompassing `edition` field in the future or we'll have only `binary_edition` or keep the distinction between `binary_edition` and a future `edition`?

ifoukarakis: IMHO that's one of the reasons we should use a string for enums (the other is to support the small number of `Unknown`). However, there doesn't seem to be a change on the editions for a long time. 

Another dimension vertical to this one is deployment type. I'd still keep them separated.

catalintomai (approved):

PR #1387, opened 2023-11-15
Chore: Add freshness checks

ifoukarakis:
#### Summary

Add freshness checks for telemetry (main rudderstack, segment and calls). Checks that all telemetry source tables have ingested events in the past 12 hours (warn) or 24 hours (error and break build if not). 

This check can be added as part of the nightly build. 

It already helped identify some source tables that can be deprecated.

### DIFF
--- a/transform/mattermost-analytics/models/staging/mattermost2/_mattermost2__sources.yml
+++ b/transform/mattermost-analytics/models/staging/mattermost2/_mattermost2__sources.yml
@@ -12,6 +12,11 @@ sources:
     tags:
       - segment
 
+    freshness: # Check that telemetry data have been received in the past 24 hours
+      warn_after: { count: 12, period: hour }
+      error_after: { count: 24, period: hour }
+    loaded_at_field: received_at
+
     tables:
       - name: activity
       - name: channel_moderation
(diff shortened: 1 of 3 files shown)


transform/mattermost-analytics/models/staging/mattermost2/_mattermost2__sources.yml
```diff
@@ -44,6 +49,7 @@ sources:
       - name: config_privacy
       - name: config_rate
       - name: config_rupport
```
catalintomai, line 51: I wonder if an easier thing to do would be to just drop the table and remove this field (and the freshness tag). I assume it is just a misspell of `config_support`.

ifoukarakis: I agree with this approach. This is what I meant by `It already helped identify some source tables that can be deprecated.`. Most of these tables have < 1k rows.

The only exception is `mattermost2.event_mobile`, which contains ~ 56 million rows, but with maximum `received_at` date equal to 2020-03-06. Since we're considering telemetry after 2020-04-01, this table can be archived. A nice side-effect is a small gain in performance.

catalintomai (approved):

PR #1388, opened 2023-11-17
Build self hosted and cloud servers spine

sri-byte:
Impact: Build self hosted and cloud servers spine


transform/mattermost-analytics/models/marts/product/licenses/_product__licenses__models.yml
```diff
@@ -91,3 +91,55 @@ models:
         description: |
           Whether the expiration date is consistent in case there's multiple sources that report the license. True if 
           more than one expiration dates are reported, false else.
+
+  - name: dim_self_hosted_customers
```
catalintomai, line 95: remove?

sri-byte: I will drop it, I'm still deciding if we need this

transform/mattermost-analytics/models/intermediate/product/licenses/_licenses_servers/int_self_hosted_licenses.sql
```diff
@@ -0,0 +1,33 @@
...
+    select distinct license_id
+        , c.portal_customer_id as customer_id
+        , c.email as email
+        , c.name as company_name
+        , coalesce(s.edition, p.name) as license_name
+        , 'Stripe' as source
+    from {{ ref('stg_stripe__subscriptions')}}  s 
+    join {{ ref('stg_stripe__customers')}}  c on s.customer_id = c.customer_id
+    left join {{ ref('stg_stripe__products')}} p on coalesce(s.product_id, s.current_product_id) = p.product_id
+    where license_id is not null
```
catalintomai, line 20: why this check is only here and not for the other two?

sri-byte: CWS doesn't contain cloud licenses. I validated this and presented findings in one of our meetings.
Legacy source data, doesn't contain `installation_id`, so a check there wasn't required.

catalintomai: this is for self-hosted `license_id is not null`

sri-byte: yes, the others don't contain any cloud data.

Stripe contains both cloud and self-hosted data hence the `license_id` filter, the check for cloud using `installation_id` field.

catalintomai: ok, so for self-hosted licenses it is ok to have null `license_ids` in the result set of the intermediate model - except for Stripe?

sri-byte: However, this is generating duplicate data. License information can be present in multiple sources causing duplicates when we join with server data. I'm trying to find a way to assign an order of precedence

transform/mattermost-analytics/models/intermediate/product/licenses/_licenses_servers/int_self_hosted_servers_spined.sql
```diff
@@ -0,0 +1,49 @@
...
+    from {{ ref('stg_mattermost2__license')}} 
+    where license_id is not null
+    group by server_id, license_id
+), spined as (
+    select ltr.server_id as server_id
+        , ltr.license_id as license_id
+        , all_days.date_day::date as activity_date
+        , {{ dbt_utils.generate_surrogate_key(['ltr.server_id', 'ltr.license_id', 'activity_date']) }} AS daily_server_license_id
+        , {{ dbt_utils.generate_surrogate_key(['ltr.server_id', 'activity_date']) }} AS daily_server_id
+        , datediff(day, ltr.min_license_date, all_days.date_day::date) as age_in_days
```
ifoukarakis, line 30: This field is not used in the final output, so it can be removed.

sri-byte:
Closing as I created a new PR along with a few changes made in this one

PR #1399, opened 2023-12-05
MM-56038 - Create mart model for calculating nps score

sri-byte:
Impact: Creates mart model for calculating nps score for instances which reported score.

#### Ticket Link
 
https://mattermost.atlassian.net/browse/MM-56038


ifoukarakis (changes requested):
I added a few comments.

There's also some other questions that need to be answered and may affect the design:
- Are the time periods for calculating NPS score fixed? I.e. are they daily, weekly, monthly, or do we need dynamic time range calculation?
- Does NPS score for each period need to be calculated at the beginning of a period or in a rolling manner?
- If a user provided a score more than once in a time period, should all scores be considered?
- Does breakdown by version need to happen on full version or major and minor part?

transform/mattermost-analytics/models/marts/product/nps/fct_nps_score.sql
```diff
@@ -0,0 +1,20 @@
+with nps_server_daily_score as (
+    select {{ dbt_utils.generate_surrogate_key(['event_date', 'server_id', 'server_version', 'user_role']) }} as score_id
+      , event_date as score_date
+      , server_id as server_id
+      , server_version as server_version
+      , user_role as user_role
+      , count(distinct case when score > 8 then user_id else null end) as promoters
+      , count(distinct case when score < 7 then user_id else null end) as detractors
+      , count(distinct case when score > 6 and score < 9 then user_id else null end) as passives
+      , count(distinct user_id) as nps_users
```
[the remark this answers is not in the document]
sri-byte: We need this `group by` expression, as mentioned in the ticket. We could do a sum of this count (of distinct users) for the selected time range within Looker. We are providing in looker this data on the granularity of date, so week and month can be rolled up.

transform/mattermost-analytics/models/marts/product/nps/fct_nps_score.sql
```diff
@@ -0,0 +1,20 @@
...
+      , round(100.0 * ((promoters / nps_users) - (detractors / nps_users)),2) as nps_score
```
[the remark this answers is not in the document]
sri-byte: Dropping this calculation so it can be calculated in Looker.

transform/mattermost-analytics/models/marts/product/nps/fct_nps_score.sql
```diff
@@ -0,0 +1,20 @@
...
+    from {{ ref('int_nps_score') }}
+    where event_date >= '{{ var('telemetry_start_date')}}'
+    group by event_date
+      , server_id 
+      , server_version 
+      , user_role
```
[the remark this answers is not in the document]
sri-byte: Updated the description

transform/mattermost-analytics/models/intermediate/product/nps/int_daily_server_nps_score.sql
```diff
@@ -0,0 +1,60 @@
...
+    LEFT JOIN
+        base_cte b ON td.date_day::date >= event_date
+    GROUP BY
+        td.date_day::date,
+        server_id,
+        user_role,
+        server_version
+    HAVING
+        td.date_day::date >= MIN(event_date)
+)
```
ifoukarakis, line 41: Is the `having` statement needed? The same filter exists on the join condition.

transform/mattermost-analytics/models/intermediate/product/nps/int_daily_server_nps_score.sql
```diff
@@ -0,0 +1,60 @@
...
+        user_role
+    ORDER BY
+        event_date DESC
+), tmp AS (
+    SELECT
+        td.date_day::date AS activity_date,
+        server_id AS server_id,
+        user_role AS user_role,
+        server_version AS server_version,
+        SUM(CASE WHEN td.date_day::date = event_date THEN promoters ELSE 0 END) AS promoters,
```
ifoukarakis, line 26: Not sure why the sum is required here. Data are already aggregated on the previous CTE. Wouldn't a `COALESCE` be enough here?

transform/mattermost-analytics/models/intermediate/product/nps/int_nps_score.sql
```diff
@@ -1,3 +1,10 @@
+{{
```
ifoukarakis, line 1: Just adding a note to remove materialization before merging.

transform/mattermost-analytics/models/intermediate/product/nps/int_daily_server_nps_score.sql
```diff
@@ -0,0 +1,60 @@
...
+        td.date_day::date,
+        server_id,
+        user_role,
+        server_version
+    HAVING
+        td.date_day::date >= MIN(event_date)
+)
+SELECT
+    tmp.*,
+    SUM(promoters) OVER (PARTITION BY server_id, user_role, server_version ORDER BY activity_date DESC ROWS BETWEEN CURRENT ROW AND 89 FOLLOWING) AS quarterly_promoters,
```
ifoukarakis, line 44: Note that `promoters` has already been aggregated before. As a result, if the same user has provided more than one score in the last 90 days, then both scores will be counted.

transform/mattermost-analytics/models/intermediate/product/nps/int_daily_server_nps_score.sql
sri-byte: This statement is reducing the total number of rows, because of the `MIN`. It is required since the `MIN` cannot in the join condition

transform/mattermost-analytics/models/intermediate/product/nps/int_daily_server_nps_score.sql
sri-byte: Without the `SUM` the code is creating 2 rows per `activity_date`, 1 row when there are promoters and another extra row with `0`'s. This is causing the quarterly calculation to break in the next CTE as I am using a window function on prior 90 days

transform/mattermost-analytics/models/intermediate/product/nps/int_daily_server_nps_score.sql
```diff
@@ -0,0 +1,60 @@
...
+        COUNT(DISTINCT user_id) AS nps_users
+    FROM
+          {{ ref('int_nps_score') }} nps_score
+    GROUP BY
+        event_date,
+        server_id,
+        server_version,
+        user_role
+    ORDER BY
+        event_date DESC
```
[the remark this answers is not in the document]
sri-byte: Removed the sort

transform/mattermost-analytics/models/intermediate/product/nps/int_user_nps_score_latest.sql
```diff
@@ -0,0 +1,10 @@
+    SELECT
```
ifoukarakis, line 1: This

transform/mattermost-analytics/models/intermediate/product/nps/int_user_nps_score_latest.sql
```diff
@@ -0,0 +1,10 @@
...
+        event_date as event_date,
+        server_id as server_id,
+        user_id as user_id,
+        server_version as server_version,
+        score as score,
+        user_role as user_role
+    FROM
+        {{ ref('int_nps_score') }} nps_score
+    qualify row_number() over (partition by server_id, user_id, server_version order by event_date desc) = 1
```
ifoukarakis, line 10: This statement will keep the last nps score for each(user, server, server version). This is different from keeping the most recent comment.

sri-byte: If they submitted more than 1 score for the same server_version, that means they went back into an old survey and submitted and updated score, this means a new survey wasn't sent since there was no version change and we update the model to the new score submitted. 

If there is a version update, a new survey will be sent and the old versions last score, and the new versions score, both will be preserved, the new score will be the recent score submitted.

sri-byte: IMO, we shouldn't have more than 1 score for any given server_version, we would be tracking more than 1 score for a version if we did that.

ifoukarakis: The new survey will be sent 21 days after the version upgrade, and only if the user hasn't completed a survey in the past 90 days. This allows for a window of the user modifying their feedback after the version upgrade.

transform/mattermost-analytics/models/marts/product/nps/fct_nps_score.sql
```diff
@@ -0,0 +1,10 @@
+{{
```
ifoukarakis, line 1: No need for this. It's materialized by default.

transform/mattermost-analytics/models/marts/product/nps/fct_nps_score.sql
```diff
@@ -0,0 +1,10 @@
...
+    config({
+        "materialized": "table",
+    })
+}}
+
+select
+    *
+from
+    {{ ref('int_daily_server_nps_score') }}
```
ifoukarakis, line 10: Perhaps remove `int_daily_server_nps_score` and push everything in this file.

ifoukarakis (approved):

transform/mattermost-analytics/models/intermediate/product/nps/int_nps_server_version_spined.sql
```diff
@@ -0,0 +1,32 @@
...
+        , COALESCE(nps_score.server_version, FIRST_VALUE(nps_score.server_version IGNORE NULLS) OVER (
+            PARTITION BY sp.server_id 
+            ORDER BY activity_date DESC
+            ROWS BETWEEN CURRENT ROW AND UNBOUNDED FOLLOWING)
+            ) AS server_version
+    from spined sp 
+    left join {{ ref('int_nps_score') }} nps_score 
+    on sp.server_id = nps_score.server_id  and sp.activity_date = nps_score.event_date
+) select * from server_version_cte 
+    order by activity_date desc
```
ifoukarakis, line 30: No need for this. The previous CTE can be the result.

transform/mattermost-analytics/models/intermediate/product/nps/int_nps_server_version_spined.sql
```diff
@@ -0,0 +1,32 @@
...
+        , server_id as server_id
+    from first_server_version fsd
+    left join {{ ref('telemetry_days') }} td 
+        on date_day::date >= min_event_date
+) 
+, server_version_cte as (
+    select 
+        sp.activity_date
+        , sp.server_id
+        , COALESCE(nps_score.server_version, FIRST_VALUE(nps_score.server_version IGNORE NULLS) OVER (
```
ifoukarakis, line 21: Why is a coalesce needed here? 3/5 it's not needed.

transform/mattermost-analytics/models/intermediate/product/nps/int_user_nps_score_spined.sql
```diff
@@ -0,0 +1,58 @@
...
+        {{ ref('telemetry_days') }} td 
+    ON 
+        date_day::date >= min_event_date
+), score_cte AS (
+    SELECT 
+        sp.activity_date,
+        sp.server_id,
+        sp.user_id,
+        nps_score.score,
+        COALESCE(nps_score.user_role, FIRST_VALUE(nps_score.user_role IGNORE NULLS) OVER (
```
ifoukarakis, line 27: Another option would be to pivot user role so that there's a protractors/detractors/passive column per each role type. This would reduce the number of rows in the final table and make aggregations easier. Also given that snowflake is columnar, the queries would be more efficient (it would eliminate the need for filtering each row).

ifoukarakis (approved):
Looks OK. One thing that might be a small source of inconsistency downsteram is that the date range for a server, role combination differs. This can be checked on servers with feedback from different roles. As an example, there's servers where total rows for `team_admin` differ from the total rows for `user` or `system_admin`.

transform/mattermost-analytics/models/intermediate/product/nps/int_user_nps_score_spined.sql
sri-byte: Sounds like a great idea, I'm all for it

transform/mattermost-analytics/models/intermediate/product/nps/int_nps_server_version_spined.sql
sri-byte: Yes, this is not required, will remove it.

transform/mattermost-analytics/models/marts/product/nps/fct_nps_score.sql
```diff
@@ -0,0 +1,85 @@
...
+SELECT a.*,
+    b.server_version AS server_version,
+    a.user_promoters + a.team_admin_promoters + a.system_admin_promoters AS promoters,
+    a.user_detractors + a.team_admin_detractors + a.system_admin_detractors AS detractors,
+    a.user_passives + a.team_admin_passives + a.system_admin_passives AS passives,
+    a.user_nps_users + a.team_admin_nps_users + a.system_admin_nps_users AS nps_users,
+    a.user_quarterly_promoters + a.team_admin_quarterly_promoters + a.system_admin_quarterly_promoters AS quarterly_promoters,
+    a.user_quarterly_detractors + a.team_admin_quarterly_detractors + a.system_admin_quarterly_detractors AS quarterly_detractors,
+    a.user_quarterly_passives + a.team_admin_quarterly_passives + a.system_admin_quarterly_passives AS quarterly_passives,
+    a.user_quarterly_nps_users + a.team_admin_quarterly_nps_users + a.system_admin_quarterly_nps_users AS quarterly_nps_users
```
ifoukarakis, line 82: nit: According to [naming conventions](https://mattermost.atlassian.net/wiki/spaces/DATAENG/pages/2422603777/DWH+Naming+Conventions), column name for aggregations should indicate the aggregation. 

Most of the columns are counts. So `promoters` could be renamed to `count_promoters_per_day`, `count_promoters_daily` or even `count_promoters`. 

Another nit: `quarterly` can be misleading and lead to confusion that it counts on current quarter (based on activity date). Perhaps use `_last_90d` or something similar as the postfix. It's already in the documentation, but it can help anyone who's doing a quick check on the fact table.

transform/mattermost-analytics/models/marts/product/nps/_nps__models.yml
```diff
@@ -37,3 +37,87 @@ models:
...
+        tests:
+          - not_null
+      - name: server_id
+        description: The ID of the server the NPS score event originated from..
+        tests:
+          - not_null
+      - name: server_version
+        description: The server's version.     
+      - name: user_role
+        description: The role of the user who submitted feedback.
```
ifoukarakis, line 59:
```suggestion
```
`user_role` column doesn't exist in table. Seems like a leftover from previous commits.

ifoukarakis (changes requested):

transform/mattermost-analytics/models/marts/product/nps/_nps__models.yml
sri-byte: Dropped this.

transform/mattermost-analytics/models/marts/product/nps/fct_nps_score.sql
sri-byte: Updated the column names following naming conventions

transform/mattermost-analytics/models/intermediate/product/nps/int_nps_server_version_spined.sql
```diff
@@ -0,0 +1,34 @@
...
+        server_id AS server_id
+    FROM first_server_version fsd
+    LEFT JOIN {{ ref('telemetry_days') }} td 
+        ON date_day::date >= min_event_date
+),
+server_version_cte AS (
+    SELECT 
+        sp.activity_date,
+        sp.server_id,
+        MAX(nps_score.server_version) AS server_version
```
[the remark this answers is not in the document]
sri-byte: I added the `max` to remove duplicates. Removed `max` now and using qualify to remove duplicates, thanks for the suggestion.

transform/mattermost-analytics/models/marts/product/nps/_nps__models.yml
```diff
@@ -37,3 +37,90 @@ models:
...
+        tests:
+          - not_null
+      - name: server_version
+        description: The server's version.
+      - name: daily_server_id
+        description: A unique identifier for the nps score generated from server_id, activity_date.
+        tests:
+          - not_null
+          - unique           
+      - name: promoters
```
ifoukarakis, line 63: Column names do not match the names in the model.

transform/mattermost-analytics/models/intermediate/product/nps/int_nps_server_version_spined.sql
```diff
@@ -0,0 +1,27 @@
+WITH first_server_version AS (
+    SELECT 
+        server_id,
+        server_version,
+        MIN(event_date) AS min_event_date
+    FROM {{ ref('int_nps_score') }}
+    WHERE event_date >= '{{ var('nps_start_date')}}'
+    GROUP BY all
```
ifoukarakis, line 8: Given that group by is happening on server_id, server_version, multiple rows for the same server_id, server_version will exist in the CTE. Shouldn't the first server date be unique for each server?

transform/mattermost-analytics/models/intermediate/product/nps/int_nps_server_version_spined.sql
```diff
@@ -0,0 +1,27 @@
...
+),
+spined AS (
+    SELECT 
+        date_day::date AS activity_date,
+        server_id AS server_id,
+        case when date_day::date = min_event_date then server_version end as server_version
+    FROM first_server_version fsd
+    LEFT JOIN {{ ref('nps_days') }} td 
+        ON date_day::date >= min_event_date
```
ifoukarakis, line 17: The previous CTE will result in multiple date/server id combinations in this CTE as well.

transform/mattermost-analytics/models/intermediate/product/nps/int_nps_server_version_spined.sql
```diff
@@ -0,0 +1,27 @@
...
+) SELECT 
+    activity_date,
+    server_id,
+    FIRST_VALUE(server_version IGNORE NULLS) OVER (
+        PARTITION BY server_id 
+        ORDER BY activity_date DESC
+        ROWS BETWEEN CURRENT ROW AND UNBOUNDED FOLLOWING
+    ) AS server_version
+FROM spined
+QUALIFY ROW_NUMBER() OVER (PARTITION BY activity_date, server_id ORDER BY activity_date DESC) = 1
```
ifoukarakis, line 27: In case of multiple versions for the same server_id, activity date, then a random one will be picked because all of them will have the same activity date. 

It would be good to be consistent about how we choose version in days with multiple versions. One approach would be to use the first or last version in a day.

transform/mattermost-analytics/models/marts/product/nps/fct_nps_score.sql
```diff
@@ -0,0 +1,94 @@
...
+  ) }}
+    FROM
+    {{ ref('int_user_nps_score_spined') }}
+    group by activity_date
+    , server_id
+)
+SELECT a.*,
+    {{ dbt_utils.generate_surrogate_key(['a.activity_date', 'a.server_id']) }} AS daily_server_id,
+    b.server_version AS server_version,
+    a.count_user_promoters_daily + a.count_team_admin_promoters_daily + a.count_system_admin_promoters_daily AS count_promoters_daily,
```
ifoukarakis, line 84: It would be nice to have a few tests for sanity checks. I.e. daily detractors + passives + promoters = count of NPS users.

ifoukarakis (changes requested):
The server spine has some issues that need to be addressed. Documentation of the fact model also needs to be updated.

transform/mattermost-analytics/models/marts/product/nps/_nps__models.yml
sri-byte: Updated yml file

transform/mattermost-analytics/models/intermediate/product/nps/int_nps_server_version_spined.sql
sri-byte: Updated this CTE, I reverted it back to the old logic. Also added qualify in the CTE where we can pick the last version for the day, using `ORDER BY timestamp desc`

transform/mattermost-analytics/models/intermediate/product/nps/int_nps_server_version_spined.sql
sri-byte: Removed the `server_version` from this and reverted it to the way it was before.

transform/mattermost-analytics/models/intermediate/product/nps/int_nps_server_version_spined.sql
sri-byte: Yes, this is now updated so this multiple combinations don't occur.

transform/mattermost-analytics/models/marts/product/nps/fct_nps_score.sql
sri-byte: Added tests to the model

transform/mattermost-analytics/models/marts/product/nps/_nps__models.yml
```diff
@@ -37,3 +37,94 @@ models:
       - name: user_email
         description: The email of the user who submitted feedback (optional).
         tags: [pii]
+
+  - name: fct_nps_score
+    tests:
+      - dbt_utils.expression_is_true:
```
ifoukarakis, line 43: nit: Tests can also have a name so that DBT output is a bit more readable. Here's how a name can be added: https://docs.getdbt.com/reference/resource-properties/data-tests#custom-data-test-name

ifoukarakis (approved):

transform/mattermost-analytics/models/intermediate/product/nps/int_user_nps_score_spined.sql
```diff
@@ -0,0 +1,65 @@
+{{
```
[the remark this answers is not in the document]
sri-byte: For the `dbt-utils.pivot` to work in the downstream model the source has to be a table, it threw an error in the CI/CD before hence the reason why I materialized as table.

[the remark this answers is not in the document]
sri-byte: I don't think the pivot will work if I remove it.

transform/mattermost-analytics/models/marts/product/nps/_nps__models.yml
sri-byte: Added names to the tests 
`count_nps_users_check_daily`
`count_nps_users_check_last90d`

PR #1410, opened 2023-12-21
Chore: Deprecate models with no matching syncs

ifoukarakis:
#### Summary

- [x] Deprecate models with no matching active syncs.
- [x] Verify that the models are not used in any Looker views (thus not visible in explores).

### DIFF
--- a/transform/snowflake-dbt/models/hightouch/freemium/professional_subscription/freemium_account.sql
+++ b/transform/snowflake-dbt/models/hightouch/freemium/professional_subscription/freemium_account.sql
@@ -1,9 +1,10 @@
-{{ config(
-    { "schema": "hightouch",
+{{
+  config({
+    "schema": "hightouch",
     "materialized": "view",
-    "tags" :["hourly","blapi"] }
-) }}
-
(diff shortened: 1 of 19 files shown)


transform/snowflake-dbt/models/hightouch/freemium/professional_subscription/freemium_account.sql
```diff
@@ -1,9 +1,10 @@
-{{ config(
-    { "schema": "hightouch",
+{{
+  config({
+    "schema": "hightouch",
     "materialized": "view",
-    "tags" :["hourly","blapi"] }
```
catalintomai, line 4: does this need "deprecated" as well?

ifoukarakis: In this PR only models used by already paused syncs have been marked as deprecated. For some (like this)  I had to add and remove the tag, as they are used by multiple syncs.

As soon as we make a decision on which other syncs are not needed anymore, a new PR can mark the models as deprecated.

catalintomai (approved):

PR #1429, opened 2024-01-16
Chore: adjust memory request size

ifoukarakis:
#### Summary

Pods are currently randomly failing due to failure to be scheduled.  This PR reduces the initial request size for the command line utilities. This is required to test whether the failure happens due to relatively big initial request size vs not enough resources on cluster.

### DIFF
--- a/airflow/dags/mattermost_dags/airflow_utils.py
+++ b/airflow/dags/mattermost_dags/airflow_utils.py
@@ -28,8 +28,8 @@ pod_defaults = {
     "cmds": ["/bin/bash", "-c"],
     "container_resources": k8s.V1ResourceRequirements(
         requests={
-            "cpu": "250m",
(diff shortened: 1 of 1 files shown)


catalintomai (approved):

catalintomai:
@ifoukarakis , would it make sense to change only one of the memory/cpu at a time to help isolate the issue?

ifoukarakis:
> @ifoukarakis , would it make sense to change only one of the memory/cpu at a time to help isolate the issue?

I noticed that it happens on the 3rd or 4th pod. All pods are currently using the same configuration. Discussed with SRE and suggested to try this configuration, just to see if pods are not scheduled due to high initial resource request.

Most of the scripts do not have high memory requirements. Let's test if we break anything by reducing initial requirements and proceed accordingly.

PR #1430, opened 2024-01-16
Chore: Adds participant_id to staging model

sri-byte:
Impact: Calls data in `RAW` tables contains `user_id` as `participant_id` in some and `actual_user_id` in other tables. The last PR which created models for calls didn't include `participant_id` which skews Active user metric calculations in Looker. This PR adds the missing field from the source tables.

### DIFF
--- a/transform/mattermost-analytics/models/staging/mm_calls_test_go/stg_mm_calls_test_go__tracks.sql
+++ b/transform/mattermost-analytics/models/staging/mm_calls_test_go/stg_mm_calls_test_go__tracks.sql
@@ -2,7 +2,7 @@ SELECT
     id AS event_id
     , event_text AS event_name
     , event AS event_table
-    , actual_user_id AS user_id
+    , coalesce(actual_user_id, participant_id) AS user_id
     , user_id AS server_id
     , received_at AS received_at
     , timestamp  AS timestamp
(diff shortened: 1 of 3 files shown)


transform/mattermost-analytics/models/staging/mm_calls_test_go/stg_mm_calls_test_go__tracks.sql
```diff
@@ -2,7 +2,7 @@ SELECT
     id AS event_id
     , event_text AS event_name
     , event AS event_table
-    , actual_user_id AS user_id
+    , coalesce(actual_user_id, participant_id) AS user_id
```
ifoukarakis, line 5: Coalescing user id with participant id makes sense only if it's the same value. A more robust design would create a lookup table of user id and possible participant ids.

sri-byte: this model is from a `base` model that unions individual event tables. The `user_id` is `actual_user_id` in some tables and `participant_id` in the other. There are no tables where it contains both fields. Because of the union we are missing the `user_id` where it is `participant_id`.

sri-byte: IMO, a look up table will make things really complex, I validated the logic against the prod staging table and the CI/CD table, and it works as expected.

ifoukarakis: Still, if `user_id` and `participant_id` have different semantics, it should be reflected, even in downstream models. Otherwise it can become easy to introduce errors downstream.

ifoukarakis (approved):
Approving based on feedback related to `participant_id`. Perhaps we can also document why we're doing this coalesce.

PR #1446, opened 2024-02-13
Chore: move focalboard_fact to nightly

ifoukarakis:
#### Summary

Move `focalboard_fact` to nightly run.


catalintomai (approved):

catalintomai:
@ifoukarakis , I think we can even move it to weekly - there is no need to get DAU/MAU every day.

PR #1451, opened 2024-02-27
MM-56793: revised version of loading data from S3 stage to snowflake

ifoukarakis:
#### Summary

Revise scripts for loading data from S3 stage to Snowflake table.

- [x] Moved references to tables/stages from hardcoded values to configuration.
- [x] Incrementally load since the last load date (or all time).
- [x] Tests.

Next step would be to create Airflow jobs for running the load.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-56793


tests/utils/push_proxy/test_service.py
```diff
@@ -0,0 +1,94 @@
...
+
+import pytest
+
+from utils.push_proxy.service import join_s3_path, load_stage_to_table
+
+
+@pytest.mark.parametrize(
+    "prefix,parts",
+    [
+        pytest.param("/path1/path2/path3", ["extra1", "extra2"], id="leading /"),
```
catalintomai, line 12: Is it guaranteed that `prefix` always start with `/`? If not maybe we can add a test case for that setup.

catalintomai (approved):

PR #1457, opened 2024-03-01
MM-55443: feature usage metrics

ifoukarakis:
#### Summary

Track feature usage per user/server.

- [x] Seed files with mapping of event to feature and a short alias.
- [x] Daily snapshot per server on the number of DAU/MAU per feature, as well as total daily and monthly (rolling) events.

Note that the events so far exist only in Rudderstack sources. No need to integrate Segment events yet.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-55443


transform/mattermost-analytics/seeds/features/event_to_feature_mapping.csv
```diff
@@ -0,0 +1,5 @@
+event_name,category,event_type,feature_name
+event,api,api_posts_ack,Message Acknowledgement (Prof)
```
catalintomai, line 2: Think you can remove the "(Prof)" (here and below). We can add it back as a separate column later.

ifoukarakis: Will remove it. It's not used anywhere right now.

catalintomai (approved):

PR #1461, opened 2024-03-11
MM-52375 Move models to nightly

sri-byte:
Impact: Move models to nightly to reduce cost as they are not being used frequently.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-52375


transform/snowflake-dbt/models/events/hourly/calls_events.sql
```diff
@@ -1,7 +1,7 @@
 {{config({
     "materialized": "incremental",
     "schema": "events",
-    "tags":"preunion"
```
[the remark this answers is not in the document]
sri-byte: `preunion` is a tag specific to hourly schedule. If we don't remove it will run in both hourly and nightly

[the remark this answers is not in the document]
sri-byte: you can see what commands are running under hourly in the settings of the job in dbt cloud

transform/snowflake-dbt/models/finance/account_daily_arr.sql
```diff
@@ -1,6 +1,7 @@
 {{config({
     "materialized": "table",
-    "schema": "finance"
+    "schema": "finance",
```
ifoukarakis, line 3: This model is also used in `marketing_funnel`, which is owned by sales. Shall we ping them?

ifoukarakis (commented):
There's quite a few models that are used downstream by `license_server_fact`. However, this model run `hourly`. I marked some of them, but perhaps we can reconsider the whole `license_server_fact` lineage as nightly.

transform/snowflake-dbt/models/events/hourly/cloud_pageview_events.sql
```diff
@@ -1,7 +1,7 @@
 {{config({
     "materialized": "incremental",
     "schema": "events",
-    "tags":"hourly"
+    "tags":["nightly"]
   })
```
catalintomai, line 5: @sri-byte , sometimes we use "[nightly]", sometimes we use "nightly" - might be good to align on one format only.

transform/snowflake-dbt/models/finance/account_daily_arr.sql
catalintomai: @ifoukarakis , `marketing_channel` sources data from `marketing_gsheets` which ingestion is turned off at the moment, so I think we are ok.

ifoukarakis (approved):

PR #1462, opened 2024-03-11
MM-57088 Update NPS Feedback categories

sri-byte:
Initial ask: Add/Update the NPS feedback categories and subcategories with new keywords

This PR aims to move away from the SQL regex solution of assigning categories & sub-categories to NPS feedback as its not maintainable. The SQL code is super sensitive and also unsure if it provides the results we are trying to achieve. 

We are trying to get away from the SQL solution by using python NLP. The comparison between manual classification, existing regex and different algorithms was shared with the team and we have agreed to proceed with the cosine similarity algorithm.

Following are the steps to push this to production.

- [x] create a package for NPS
- [ ] clean up the code
- [ ] add tests
- [ ] integrate with airflow
- [ ] document the workflow in Confluence
- [ ] integrate with NPS models

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57088


ifoukarakis:
Closing as incomplete. CC @catalintomai

PR #1481, opened 2024-04-11
Chore: add customer information dimension

ifoukarakis:
#### Summary

Reuse code from active user base analysis to create a dimension with known customer information.


transform/mattermost-analytics/analyses/active_user_base.sql
```diff
@@ -74,7 +74,7 @@ rudder_licenses as (
 ), license_data as (
     -- Gather all self-hosted license data from CWS and legacy licenses.
     select
-        license_id
+        license_id.
```
catalintomai, line 77: intentional?

transform/mattermost-analytics/models/marts/product/licenses/dim_latest_server_customer_info.sql
```diff
@@ -0,0 +1,18 @@
+select
+    server_id
+    , license_id
+    , installation_id
+    , coalesce(license_company_name, cloud_company_name) as company_name
+    , coalesce(license_contact_email, cloud_contact_email) as contact_email
+    , coalesce(license_sku, cloud_sku) as sku
+    , license_expire_at as license_expire_at
```
catalintomai, line 8: nit: remove the aliasing here and below since we don't do it in other places (e.g. ast_license_telemetry_date or others).

transform/mattermost-analytics/models/marts/product/licenses/_product__licenses__models.yml
```diff
@@ -142,4 +142,59 @@ models:
...
+        description: | 
+          Boolean flag, indicating whether license data have been found for the current license id. 
+          - True if matching data have been found.
+          - False if no license id has been found or if no matching license exists.
+
+      - name: found_matching_stripe_entry
+        description: | 
+          Boolean flag, indicating whether data for the current installation ID are available in Stripe. 
+          - True if matching data have been found.
+          - False if no data have been found or if there's no installation ID.
```
catalintomai, line 196: "no data have been found or" condition below checks for `installation_id not null" only

transform/mattermost-analytics/analyses/active_user_base.sql
ifoukarakis: Good catch. Was using this file as a reference and seems this is a miss-type.

catalintomai (approved):

PR #1509, opened 2024-05-14
Add subscription history model

nickmisasi:
#### Summary
Adds a model for subscription history to aid in building out a "Cloud 360 Dashboard" - this specifically aids in exposing the current "billable" seat count on the workspace.

### DIFF
--- a/transform/mattermost-analytics/models/marts/sales/_sales__models.yml
+++ b/transform/mattermost-analytics/models/marts/sales/_sales__models.yml
@@ -1,0 +1,17 @@
+version: 2
+
+
+models:
+  - name: fct_subscription_history
+    description: |
+      Cloud subscription seat history data.
+    columns:
+    - name: subscription_history_id
+      description: The subscription history's ID.
+    - name: subscription_id
+      description: The subscription's ID.
+    - name: licensed_seats
+      description: The number of licensed seats active at the time of this event.
+    - name: created_at
+      description: The creation timestamp of this subscription history event
+
--- a/transform/mattermost-analytics/models/marts/sales/fct_subscription_history.sql
+++ b/transform/mattermost-analytics/models/marts/sales/fct_subscription_history.sql
@@ -1,0 +1,11 @@
+with subscription_history as (
+    select 
+        subscription_history_event_id
+        ,  subscription_id
+        ,  licensed_seats
+        , created_at
+
+    from {{ ref('stg_cws__subscription_history') }}
+)
+
+select * from subscription_history
--- a/transform/mattermost-analytics/models/staging/cws/_cws__models.yml
+++ b/transform/mattermost-analytics/models/staging/cws/_cws__models.yml
@@ -159,3 +159,20 @@ models:
         description: Unique Identifier for licenses.
       - name: trial_request_id
         description: The trial request's ID.
+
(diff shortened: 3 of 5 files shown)


transform/mattermost-analytics/models/marts/sales/_sales__models.yml
```diff
@@ -0,0 +1,17 @@
+version: 2
+
+
+models:
+  - name: fct_subscription_history
+    description: |
+      Cloud subscription seat history data.
+    columns:
+    - name: subscription_history_id
```
ifoukarakis, line 9: It's a common practice to create a unique key for each row in fact tables. If the key is a combination of columns, then a surrogate key can be generated from these columns:

Here's an example: https://github.com/mattermost/mattermost-data-warehouse/blob/1a8f222effc50d1131f3377649da85c3fd473d35/transform/mattermost-analytics/models/marts/product/fct_active_users.sql#L16.

transform/mattermost-analytics/models/marts/sales/_sales__models.yml
```diff
@@ -0,0 +1,17 @@
+version: 2
+
+
+models:
+  - name: fct_subscription_history
```
ifoukarakis, line 5: If I understand correctly, the intention is to join these data with stripe data. Perhaps the join can happen here so that data become available in a form that's easier to handle in Looker.

transform/mattermost-analytics/models/marts/sales/_sales__models.yml
nickmisasi: The `subscription_history_id` field is a unique ID assigned to a subscription history event by the CWS - it's an alphanumeric 26 character string. I don't think we need a surrogate key given this?

transform/mattermost-analytics/models/marts/sales/_sales__models.yml
nickmisasi: Do you have an example in mind of what this might look like? I don't see another fct model in the marts folder for Stripe - I also want to note that this may not only join against Subscription data, but with customer data, invoice data, etc - we don't actually know what'll be necessary in the future. I'm not sure the ramifications of leaving it vague or being more specific here - is it mostly performance?

ifoukarakis: Mostly performance and complexity. Adding complex join logic in looker can cause issues in both. The idea is that the models in the mart layer are following the star schema conventions (or similar). Short story is that a fact table contains either numeric/boolean values or keys to dimension tables. Dimension tables contain keys and anything that can be used to filter the fact table by. There's a ton of details here that depend on the use case. 

Since it's not possible to predict what's coming in the future, we can start with the bare minimum for the current use case. Future PRs can extend existing fact/dimension tables or introduce new ones as needed.

ifoukarakis (approved):

transform/mattermost-analytics/models/marts/sales/_sales__models.yml
nickmisasi: I think that's a good plan - once I've really narrowed down what the dashboard will look like I can definitely open a subsequent PR to address perf concerns. Thanks!

PR #1525, opened 2024-05-29
Chore: add installation id summary

ifoukarakis:
#### Summary

Add latest summary for each installation ID.

### DIFF
--- a/transform/mattermost-analytics/models/marts/product/_product__models.yml
+++ b/transform/mattermost-analytics/models/marts/product/_product__models.yml
@@ -356,3 +356,47 @@ models:
           Total number of active users for the past 24 hours prior to the timestamp of the event.
           Bots and deleted users are excluded. Reported by mattermost server. If there is no 
           information about the specific date (i.e. server inactive or server hasn't reported telemetry), then it's 0.
+
+  - name: dim_installation_summary
+    description: | 
+      Summary of each installation based on the telemetry data. Contains data for servers with an existing 
+      installation id only.
+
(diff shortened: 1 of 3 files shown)


transform/mattermost-analytics/models/marts/product/_product__models.yml
```diff
@@ -355,4 +355,48 @@ models:
...
+      Summary of each installation based on the telemetry data. Contains data for servers with an existing 
+      installation id only.
+
+    columns:
+      - name: installation_id
+        description: The id of the installation.
+        tests:
+          - not_null
+          - unique
+      - name: installation_id
```
catalintomai, line 371: duplicate?

ifoukarakis: Hah, seems I accidentally added that using Github Copilot. Nice catch!

catalintomai: Next step - adding comments (like this one) to PRs = )

ifoukarakis:
Note that tests are failing due to installations ids existing in stripe but not in telemetry. We need to decide what to do with them.

nickmisasi (approved):
LGTM!

ifoukarakis:
> Note that tests are failing due to installations ids existing in stripe but not in telemetry. We need to decide what to do with them.

Following discussion with @nickmisasi, there are reasons it would happen, but safe to ignore them. Removed check for now.

PR #1535, opened 2024-06-18
MM-58699 - Change type for is_enterprise_ready field

catalintomai:
#### Summary
Change type for `is_enterprise_ready` field 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58699

### DIFF
--- a/transform/mattermost-analytics/models/intermediate/product/servers/int_server_active_days_spined.sql
+++ b/transform/mattermost-analytics/models/intermediate/product/servers/int_server_active_days_spined.sql
@@ -75,7 +75,10 @@ select
     coalesce(t.operating_system, l.operating_system, d.operating_system) as operating_system,
     coalesce(t.database_type, l.database_type, d.database_type) as database_type,
     coalesce(t.database_version, l.database_version) as database_version,
-    coalesce(t.edition, l.edition, d.is_enterprise_ready) as is_enterprise_ready,
+    case
+        when coalesce(t.edition, l.edition, d.is_enterprise_ready) = true then true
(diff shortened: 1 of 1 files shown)


transform/mattermost-analytics/models/intermediate/product/servers/int_server_active_days_spined.sql
```diff
@@ -75,7 +75,10 @@ select
     coalesce(t.operating_system, l.operating_system, d.operating_system) as operating_system,
     coalesce(t.database_type, l.database_type, d.database_type) as database_type,
     coalesce(t.database_version, l.database_version) as database_version,
-    coalesce(t.edition, l.edition, d.is_enterprise_ready) as is_enterprise_ready,
+    case
+        when coalesce(t.edition, l.edition, d.is_enterprise_ready) = true then true
+        else false
+    end as is_enterprise_ready,
```
ifoukarakis, line 81: If edition or `is_enterperise_ready` is null, then the resulting value will be null. Is this something we can safely assume?

ifoukarakis: After having a closer look, I noticed that there's an inconsistency in the data types of each column. The first two values (coming from rudderstack - `t` - and segment - `l`) are actually sent as strings rather than booleans. This is also visible in [MM code](https://github.com/mattermost/mattermost/blob/master/server/platform/services/telemetry/telemetry.go#L1501).

A better solution would be to convert these values to booleans in the staging models.

PR #1541, opened 2024-06-27
Chore: simplify filtering in downloads

ifoukarakis:
#### Summary

- [x] Remove filtering based on download size.
- [x] Add filtering based on HTTP method.


transform/mattermost-analytics/models/marts/product/downloads/fct_downloads.sql
```diff
@@ -38,16 +38,15 @@ select
 
 from
     {{ ref('stg_releases__log_entries') }} le
-    left join {{ ref('int_download_stats_per_uri') }} ds on le.uri = ds.uri
 where
     -- Keep only requests with responses at least 1 mb
     response_bytes > 1000000
```
catalintomai, line 43: I guess this assumes that in the case of a chunked download the minimum chunk size is 1MB(?)

catalintomai (approved):

transform/mattermost-analytics/models/marts/product/downloads/fct_downloads.sql
ifoukarakis: TBH this was a filter in the current implementation, so I kept it in order to not have big differences. Let's see the difference between this branch and the current version and we can decide whether to keep it or not.

PR #1563, opened 2024-07-24
Hotfix: properly handle different user id columns

ifoukarakis:
#### Summary

Build on top of previous hotfix. Seems that some events use `user_actual_id`, while others use `actual_user_id` for user id.


catalintomai (approved):

catalintomai:
@ifoukarakis - should we open a bug on the Copilot team to fix this?

ifoukarakis:
> @ifoukarakis - should we open a bug on the Copilot team to fix this anyway (and before there are more versions having the issue) ?

Yes, I'll open one.

PR #1568, opened 2024-08-06
MM-60014 - Update freshness check on `user_start_recording` Calls event.

catalintomai:
#### Summary
Update freshness check on `user_start_recording` Calls event.
Note: this unblocks current dbt mattermost-analytics nightly job failures.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60014

### DIFF
--- a/transform/mattermost-analytics/models/staging/mm_calls_test_go/_mm_calls_test_go__sources.yml
+++ b/transform/mattermost-analytics/models/staging/mm_calls_test_go/_mm_calls_test_go__sources.yml
@@ -34,6 +34,7 @@ sources:
       - name: user_raise_hand
       - name: user_share_screen
       - name: user_start_recording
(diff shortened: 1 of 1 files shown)


ifoukarakis (approved):
Another alternative would be to change this to 7 days for `user_start_recording` (the current max date diff between two subsequent events) and 14 days for `user_stop_recording`.

catalintomai:
"Another alternative would be to change this to 7 days for user_start_recording (the current max date diff between two subsequent events) and 14 days for user_stop_recording." - I like it better (also matches better my discussion with Claudio on the issues - he mentioned recording events being more common mid-week). 
re: actual way of achieving it - I assume table freshness settings supersede source freshness setting, but not sure about the right way to do it here, so experimenting a bit.

ifoukarakis (approved):

PR #1571, opened 2024-08-08
Hotfix: fix duplicate ids in grp_performance_events

ifoukarakis:
#### Summary

Fix appending same event more than once.

### DIFF
--- a/transform/mattermost-analytics/models/marts/web_app/platform/grp_performance_events.sql
+++ b/transform/mattermost-analytics/models/marts/web_app/platform/grp_performance_events.sql
@@ -2,7 +2,8 @@
     config({
         "materialized": "incremental",
         "cluster_by": ['event_date'],
-        "incremental_strategy": "append",
+        "unique_key": 'id',
+        "incremental_strategy": "delete+insert",
         "snowflake_warehouse": "transform_l"
     })
 }}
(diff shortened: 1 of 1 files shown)


transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_prod_performance_events.sql
```diff
@@ -11,12 +11,8 @@ SELECT
...
     {{ ref('stg_mm_telemetry_prod__performance_events') }}
 WHERE
-    id not in (
-        'd\';waitfor/**/delay\'0:0:0\'/**/--/**/',
-        's\');waitfor/**/delay\'0:0:0\'/**/--/**/',
-        'v\';waitfor/**/delay\'0:0:2\'/**/--/**/',
-        'o\');waitfor/**/delay\'0:0:0\'/**/--/**/'
-    )
+    -- Exclude non-UUID strings
+    not id ilike '%waitfor%'
```
catalintomai, line 15: there could be more ids that are having some invalid strings in them other than waitfor - wondering if we should open a ticket to expand this check to make it more general (and not only for perf events).

catalintomai (approved):

transform/mattermost-analytics/models/intermediate/web_app/platform/int_mm_telemetry_prod_performance_events.sql
ifoukarakis: We can replace it with a regex to check if id is a UUID. But this could break if rudderstack decides to change the format of id. 

Given the frequency of these issues, I think this change should be enough for now.

PR #1573, opened 2024-08-13
MM-60178 - propagate more Stripe data source fields to the fct_trial_request_history table

catalintomai:
#### Summary
Propagate more Stripe data source fields to the `fct_trial_request_history` table.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60178

### DIFF
--- a/transform/mattermost-analytics/models/marts/product/trial_requests/fct_trial_request_history.sql
+++ b/transform/mattermost-analytics/models/marts/product/trial_requests/fct_trial_request_history.sql
@@ -40,7 +40,7 @@ select
     , tr.end_at
     , tr.request_source
     , tr.request_type
-    , tr.product_id
+    , tr.stripe_product_id
     , tr.converted_to_paid_at
     , tr.status
     , agg.first_trial_start_at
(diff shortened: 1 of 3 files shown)


transform/mattermost-analytics/models/marts/product/trial_requests/fct_trial_request_history.sql
```diff
@@ -40,6 +40,9 @@ select
     , tr.end_at
     , tr.request_source
     , tr.request_type
+    , tr.product_id
```
ifoukarakis, line 43: Shall we create a `dim_products`? I'm a bit uncertain as we'll need to figure out sources outside Stripe and correlate them. This will mean that a separate ID will need to be added.

ifoukarakis, line 43: For now, perhaps we can rename this to `stripe_product_id` to be more explicit about the origin of this product.

ifoukarakis (approved):

transform/mattermost-analytics/models/marts/product/trial_requests/fct_trial_request_history.sql
catalintomai: I think it make sense eventually - I see Stripe, CWS (though still calles stripe_product_id) and SFDC as current sources. For now we can use this one and revisit once we hit in more places.

ifoukarakis (approved):

PR #1582, opened 2024-08-20
MM-58167: add country to server information

ifoukarakis:
#### Summary

- [x] Add latest country info to each server summary.
- [x] Add country exclusion reason.

Some of the logic encapsulated in `dim_server_info` was moved to intermediate layer models in order to support both adding country exclusion reason and adding country in `dim_server_info`. This is an optimization that helps in running the heavy join query between server ip and country. It also helps by pruning the join size due to performing the range join on distinct ips (~ 33% of full size).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58167

### DIFF
--- a/transform/mattermost-analytics/models/marts/product/dim_server_info.sql
+++ b/transform/mattermost-analytics/models/marts/product/dim_server_info.sql
@@ -1,20 +1,4 @@
-with server_telemetry_summary as (
-    select
-        server_id,
-        min(activity_date) over (partition by server_id) as first_activity_date,
-        max(activity_date) over (partition by server_id ) as last_activity_date,
-        first_value(binary_edition) over (partition by server_id order by activity_date asc) as first_binary_edition,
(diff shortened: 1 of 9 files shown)


transform/mattermost-analytics/models/marts/product/dim_server_info.sql
```diff
@@ -59,7 +45,10 @@ select
     si.first_count_registered_active_users,
     si.last_count_registered_active_users,
     si.last_daily_active_users,
-    si.last_monthly_active_users
+    si.last_monthly_active_users,
+    si.last_server_ip,
+    coalesce(ip.last_known_ip_country, 'Unknown') as last_known_ip_country
```
catalintomai, line 50: can this (ever) happen given the similar code in int_server_ip_to_country.sql?

ifoukarakis: Shouldn't be needed, but added it as an extra safeguard.

catalintomai (approved):

PR #1624, opened 2024-10-07
Chore: surface shared channels settings

catalintomai:
#### Summary
Surface shared channels settings to the active servers model.


ifoukarakis (approved):
Approving, but it would be nice to also update `_product__models.sql`.

PR #1628, opened 2024-10-08
Chore: remove dim_*_customers models

catalintomai:
#### Summary
Remove dim_*_customers models.

https://github.com/mattermost/mattermost-looker/pull/1056 needs to be checked-in before this one.


ifoukarakis (commented):
We can remove the models as well.

ifoukarakis (approved):

PR #1629, opened 2024-10-09
MM-60769: load from s3

ifoukarakis:
#### Summary

- [x] Load packets from S3.
- [x] Support incremental loading. 

Assumes that different types of packets are under `<prefix>/YYYY/MM/DD/`.

#### Ticket Link

https://mattermost.atlassian.net/issues/MM-60769

### DIFF
--- a/utils/packets/__main__.py
+++ b/utils/packets/__main__.py
@@ -5,7 +5,12 @@ import click
 
 from utils.db.helpers import snowflake_engine
 from utils.helpers import initialize_cli_logging
-from utils.packets.service import ingest_support_packet, ingest_survey_packet
+from utils.packets.service import (
+    ingest_support_packet,
+    ingest_support_packets_from_s3,
+    ingest_survey_packet,
+    ingest_surveys_from_s3,
+)
 
 initialize_cli_logging(logging.INFO, 'stderr')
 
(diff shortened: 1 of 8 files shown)


utils/packets/aws.py
```diff
@@ -0,0 +1,37 @@
...
+    Check if a bucket exists.
+    """
+    try:
+        s3.head_bucket(Bucket=bucket)
+        return True
+    except ClientError:
+        return False
+
+
+# The bucket does not exist or you ha
```
catalintomai, line 21: text cut?

utils/packets/__main__.py
```diff
@@ -5,7 +5,12 @@
 
 from utils.db.helpers import snowflake_engine
 from utils.helpers import initialize_cli_logging
-from utils.packets.service import ingest_support_packet, ingest_survey_packet
+from utils.packets.service import (
+    ingest_support_packet,
+    ingest_support_packets_from_s3,
```
catalintomai, line 10: I assume there is no need to add tests for `ingest_support_packets_from_s3` (only the table name changes).

catalintomai (approved):

utils/packets/aws.py
ifoukarakis: Left over. Removing.

utils/packets/__main__.py
ifoukarakis: It's calling the same function as `ingest_surveys_from_s3`. The individual components have been tested. We can add an end-to-end test if you prefer. However, since there might be changes depending on the structure of the S3 bucket, I prefer to wait for them before adding the extra data.

PR #1630, opened 2024-10-09
Chore: surface full server version info to the nps models

catalintomai:
#### Summary
Surface full server version info to the nps models. Replace the previously code-built server_version. Generate version_id to be able to join with `dim_version` later.


ifoukarakis (approved):

catalintomai:
@ifoukarakis , will remove the table materialization from `int_nps_server_version_spined.sql` before checking in.

PR #1631, opened 2024-10-10
MM-60600: CLI tools for deduplicated event table

ifoukarakis:
#### Summary

- [x] CLI utils for deferred merge.
- [x] Handle schema changes when applying merge.
- [x] Minor refactoring to avoid duplicate code.
- [x] DBT models for maintaining delta table and deduped view.  

Doesn't include:
- Replacing current model with new deduplicated one.
- Automation 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60600


transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__event_deduped.sql
```diff
@@ -0,0 +1,319 @@
...
+        , num_of_request
+        , num_total
+        , org_id
+        , original_timestamp
+        , persistent_notifications
+        , playbooks
+        , plugin_id
+        , post_id
+        , privacy
+        , q
```
catalintomai, line 254: might be worth removing the spurious columns before running the process.

ifoukarakis: We should check values on these columns first. Let's do it in a separate PR in order to keep things simpler.

PR #1639, opened 2024-10-24
Chore: Surface license settings

catalintomai:
#### Summary
Surface more license settings.


transform/mattermost-analytics/models/marts/product/fct_active_servers.sql
```diff
@@ -26,15 +25,12 @@ select
...
     -- Metadata
     s.has_telemetry_data,
     s.has_legacy_telemetry_data,
     s.has_diagnostics_data,
     s.is_missing_activity_data,
-    l.is_feature_shared_channels_enabled,
-    l.is_feature_remote_cluster_service_enabled,
     l.daily_server_id is null as is_missing_license_data
 from
     {{ ref('int_server_active_days_spined') }} s
```
catalintomai, line 36: Note: kept join with `int_server_license_daily` because it is is used in `_product__models.yml`, line #250. If we can get rid of that, then we can safely remove it here as well.
`

ifoukarakis: It might be possible to replace it with a referential integrity check. Let's do this in a separate PR.

ifoukarakis (approved):

PR #1642, opened 2024-10-25
Chore: add daily_server_id field to nps_feedback model

catalintomai:
#### Summary
Add `daily_server_id` field to `nps_feedback` model. Needed later on in Looker to be able to join with `dim_daily_license` to surface the company name (and other info).


transform/mattermost-analytics/models/marts/product/nps/fct_nps_feedback.sql
```diff
@@ -1,4 +1,5 @@
 select {{ dbt_utils.generate_surrogate_key(['nf.server_id', 'nf.user_id', 'nf.feedback', 'nf.event_date']) }} as nps_feedback_id
+       , {{ dbt_utils.generate_surrogate_key(['nf.event_date', 'nf.server_id']) }} AS daily_server_id,
```
ifoukarakis: Order of fields seems to be different in the majority of the models (`server_id` then date column). Seems there's a bug in `fct_nps_score` as well. Let's use the following suggestion here: 

```suggestion
       , {{ dbt_utils.generate_surrogate_key(['nf.server_id', 'nf.event_date']) }} AS daily_server_id,
```

ifoukarakis (changes requested):

ifoukarakis (approved):

PR #1651, opened 2024-10-31
MM-61128: Add support for incremental config server data models

catalintomai:
#### Summary
Add support for incremental config server data models.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61128

### DIFF
--- a/transform/mattermost-analytics/models/intermediate/product/configs/int__config_ldap.sql
+++ b/transform/mattermost-analytics/models/intermediate/product/configs/int__config_ldap.sql
@@ -1,0 +1,91 @@
+select event_id
+    , event_table
+    , event_name
+    , server_id
+    , received_at
+    , timestamp
+
+    , connection_security_ldap
+    , enable_ldap
+    , enable_admin_filter
+    , enable_sync
+    , isdefault_email_attribute_ldap
+    , isdefault_first_name_attribute_ldap
+    , isdefault_group_display_name_attribute
+    , isdefault_group_id_attribute
+    , isdefault_id_attribute_ldap
+    , isdefault_last_name_attribute_ldap
+    , isdefault_login_button_border_color_ldap
+    , isdefault_login_button_color_ldap
+    , isdefault_login_button_text_color_ldap
+    , isdefault_login_field_name
+    , isdefault_login_id_attribute
+    , isdefault_nickname_attribute_ldap
+    , isdefault_position_attribute_ldap
+    , isdefault_username_attribute_ldap
+    , isempty_admin_filter
+    , isempty_group_filter
+    , isempty_guest_filter
+    , isnotempty_picture_attribute
+    , null as isnotempty_private_key
+    , null as isnotempty_public_certificate
+    , max_page_size
+    , query_timeout_ldap
+    , segment_dedupe_id_ldap
+    , skip_certificate_verification
+    , sync_interval_minutes
+
+    -- Metadata from Segment
+    , context_library_name
+    , context_library_version
+    , sent_at
+    , original_timestamp
+from {{ ref('int_mattermost2__config_ldap') }}
+union
+select event_id
+    , event_table
+    , event_name
+    , server_id
+    , received_at
+    , timestamp
+
+    , installation_id
+    , anonymous_id
+    , server_ip
+
+    , connection_security_ldap
+    , enable_ldap
+    , enable_admin_filter
+    , enable_sync
+    , isdefault_email_attribute_ldap
+    , isdefault_first_name_attribute_ldap
+    , isdefault_group_display_name_attribute
+    , isdefault_group_id_attribute
+    , isdefault_id_attribute_ldap
+    , isdefault_last_name_attribute_ldap
+    , isdefault_login_button_border_color_ldap
+    , isdefault_login_button_color_ldap
+    , isdefault_login_button_text_color_ldap
+    , isdefault_login_field_name
+    , isdefault_login_id_attribute
+    , isdefault_nickname_attribute_ldap
+    , isdefault_position_attribute_ldap
+    , isdefault_username_attribute_ldap
+    , isempty_admin_filter
+    , isempty_group_filter
+    , isempty_guest_filter
+    , isnotempty_picture_attribute
+    , isnotempty_private_key
+    , isnotempty_public_certificate
+    , max_page_size
+    , query_timeout_ldap
+    , null as segment_dedupe_id_ldap
+    , skip_certificate_verification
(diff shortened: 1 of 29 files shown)


transform/mattermost-analytics/models/intermediate/product/configs/int__config_ldap.sql
```diff
@@ -0,0 +1,101 @@
+{{
+    config({
```
catalintomai, line 2: These will be removed before checkin.

transform/mattermost-analytics/models/intermediate/product/configs/int_mattermost2__config_ldap.sql
```diff
@@ -0,0 +1,22 @@
...
+}}
+
+select 
+     {{ dbt_utils.star(ref('stg_mattermost2__ldap')) }}
+     , cast(received_at AS date) AS received_at_date
+     , server_id is not null as has_segment_telemetry_data
+    from {{ ref('stg_mattermost2__ldap') }} tmp
+    where
+    -- Exclude items with missing timestamps
+       timestamp is not null
```
ifoukarakis, line 18: Since this is injected by Rudder/segment if missing, it shouldn't be null. Are there any example cases that this happened? Otherwise we can remove it.

transform/mattermost-analytics/models/intermediate/product/configs/int__config_ldap.sql
```diff
@@ -0,0 +1,101 @@
...
+        "materialized": "table",
+        "snowflake_warehouse": "transform_xs"
+    })
+}}
+
+select event_id
```
ifoukarakis, line 8: An alternative would be to use [dbt_utils.union_relations](https://github.com/dbt-labs/dbt-utils?tab=readme-ov-file#union_relations-source). It would preserve columns and make the code a bit more readable.

transform/mattermost-analytics/models/intermediate/product/configs/int__config_ldap.sql
```diff
@@ -0,0 +1,101 @@
+{{
+    config({
+        "materialized": "table",
+        "snowflake_warehouse": "transform_xs"
```
ifoukarakis, line 4: These models took significant time to run on xs warehouse. Perhaps we can test them on a large. However we'll need a clean environment in order to avoid running them in incremental mode.

transform/mattermost-analytics/models/marts/product/dim_daily_server_config.sql
```diff
@@ -1,496 +1,352 @@
 {{ config(
     tags=["disabled"]
```
ifoukarakis, line 2: Let's remove this tag from this model in order for the model to run on this branch. This will help get an estimate of how long the first run will take.

ifoukarakis: Also let's make sure that this is explicitly running on a large warehouse. XS might look cheaper, but spilling to disk is likely to be causing longer run times.

transform/mattermost-analytics/models/marts/product/dim_daily_server_config.sql
```diff
@@ -1,496 +1,355 @@
...
-    tags=["disabled"]
-) }}
-with segment_oauth as (
-    select
-        server_id
-        , cast(timestamp as date) as server_date
-        , {{ dbt_utils.generate_surrogate_key(['server_id', 'server_date']) }} as daily_server_id
+{{
+    config({
+        "materialized": "table",
```
catalintomai, line 3: will remove this line before checkin.

transform/mattermost-analytics/models/intermediate/product/configs/int__config_plugin.sql
```diff
@@ -0,0 +1,323 @@
...
+    , installation_id
+    , anonymous_id
+    , server_ip
+    , allow_insecure_download_url
+    , automatic_prepackaged_plugins
+    , chimera_oauth_proxy_url
+    , enable_plugin
+    , enable_alertmanager
+    , enable_antivirus
+    , coalesce(enable_autolink,
```
ifoukarakis, line 141: These coalesce expressions can be moved to the staging model. Then `dbt_utils.union_relations` can be used in this model.

transform/mattermost-analytics/models/intermediate/product/configs/int_mm_telemetry_prod__config_plugin.sql
```diff
@@ -0,0 +1,21 @@
...
+        "incremental_strategy": "delete+insert",
+        "cluster_by": ['received_at_date'],
+        "snowflake_warehouse": "transform_l"
+    })
+}}
+
+select 
+     {{ dbt_utils.star(ref('stg_mm_telemetry_prod__plugin')) }}
+     , cast(received_at AS date) AS received_at_date
+     , server_id is not null as has_rudderstack_telemetry_data
```
ifoukarakis, line 14: nit: Given that this is a rudderstack source, it can be changed to 
```suggestion
     , true as has_rudderstack_telemetry_data
```

Same goes for segment.

ifoukarakis: 4/5 this information might get lost because of the `qualify` expression later in the `dim_daily_server_config`. The `qualify` expression will keep the flag only for the latest record. In most cases this will be OK, since servers usually are either in a really old or a new version. 3/5 the only exception is if a server is upgraded from an old version to a newer one and rolled back.

ifoukarakis (commented):
Left a comment for `int__config_plugin`. Also perhaps we should try to run it one more time to monitor timings for the incremental models.

Otherwise LGTM.

transform/mattermost-analytics/models/intermediate/product/configs/int_mm_telemetry_prod__config_plugin.sql
catalintomai: I think we can live with the corner-case, at least for now.

ifoukarakis (approved):
LGTM!

transform/mattermost-analytics/models/intermediate/product/configs/int__config_ldap.sql
catalintomai: cc: @ifoukarakis , actually I take that back. Because of the errors here: https://cloud.getdbt.com/deploy/13604/projects/216476/runs/340761541/ , I don't think I can remove the materialized statement.

PR #1661, opened 2024-11-06
Model support for consolidation of server events;

catalintomai:
#### Summary
Add preliminary changes to ingest aggregated server config data into the DWH. 
### Dependencies
https://github.com/mattermost/mattermost/pull/28200


transform/mattermost-analytics/models/intermediate/product/configs/int__config_all.sql
```diff
@@ -0,0 +1,15 @@
+{{
```
ifoukarakis, line 1: Is this needed? It's a union of a single relation.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__all.sql
```diff
@@ -0,0 +1,255 @@
+
+with source as (
```
ifoukarakis, line 2: The name of the model is too generic. Perhaps it can be renamed to `stg_mm_telemetry_prod__configs` or `stg_mm_telemetry_prod__consolidated_configs`.

transform/mattermost-analytics/models/staging/mm_telemetry_prod/stg_mm_telemetry_prod__all.sql
```diff
@@ -0,0 +1,255 @@
...
+        , config_oauth_enable_google                          as is_google_enabled
+        , config_oauth_enable_office_365                      as is_office365_enabled
+        , config_oauth_enable_openid                          as is_openid_enabled
+        , config_oauth_openid_gitlab                          as is_openid_gitlab_enabled
+        , config_oauth_openid_google                          as is_openid_google_enabled
+        , config_oauth_openid_office_365                      as is_openid_office365_enabled
+
+        -- Ldap information
+        , config_ldap_connection_security                     as connection_security_ldap
+        , config_ldap_enable                                  as enable_ldap
```
ifoukarakis, line 31: 1/5 Should the `_ldap` prefix kept in all columns?

catalintomai: I kept the naming from `stg_mm_telemetry_prod__ldap.sql` (which I believe was based on the old repo naming). If we change it here we'll have to change it there as well. I say, let's keep it as it is and if we see issues/confusions we can change later.

ifoukarakis (approved):

PR #1681, opened 2024-12-06
Chore: Deprecate model that is not used anymore

ifoukarakis:
#### Summary

Deprecate model that is not useful anymore.


catalintomai (approved):

ifoukarakis:
Part of the PR is needed in downstream models. Skipping deprecation for now.

